PR #27, opened 2026-05-05
Fix tombstone purge propagation, mid-sync write races, and remote listing latency (2026.0505.1)

VertigoRay:
## Summary

- **Tombstone correctness** — reduces `tombstoneJitterMs` default from 30s to 5s; adds migration shim for existing installs; hardens the mtime gate against malformed `deleted_at` values (non-finite, zero, negative, implausibly future); adds a shared `_cleared.json` marker so `keep-local-purge-tombstone` events propagate across devices and peer shards stop honoring stale tombstones
- **Mid-sync write protection** — records `syncStartTs` at the top of each cycle; files modified by the user during the sync window are not recorded as cleanly synced; re-checks local mtime before download overwrite and skips if the file changed (respects `conflictStrategy`)
- **Remote listing performance** — replaces sequential directory BFS with `Promise.allSettled` batches of 5; failed folder listings are logged and skipped instead of aborting the scan

## Verification

```
npm ci
npx jest --runInBand   →  108 passed (6 suites)
npm run build          →  main.js 39.2 KB
```

## Not included

- Remote mtime cache (measure first after this PR)
- Vault event listeners / pending-change queue (future architecture work)
- Removal of second `getLocalFiles()` scan (safe only after broader in-place state model)

## Test plan

- [ ] Install the plugin (or update via BRAT) and confirm settings load without console errors
- [ ] Verify `tombstoneJitterMs` in plugin data resets to 5000 if it was 30000
- [ ] Delete a file on Device A, sync, sync Device B — confirm file does not reappear
- [ ] Edit a file on Device A during an active sync from Device B — confirm edit is not overwritten
- [ ] Observe sync time improvement on a vault with many nested folders


VertigoRay:
Reopening on ForkTheGhost fork.

PR #92, opened 2026-05-22
Harden mobile sync follow-ups and config excludes

VertigoRay:
Summary
- replaces the debug modal bottom Close button with Copy snippet, copying the latest RUNTIME line plus the last 5 log lines
- centralizes notes-first Obsidian config excludes and applies gitignore-style matching across native Git, desktop Git-over-File-Station, and mobile Git-over-File-Station
- fixes mobile repeat sync after first materialization so established checkouts are not re-preserved as hundreds of conflict copies

Context
- v2026.0522.9 successfully materialized the mobile vault, but the next scheduled sync treated the already-materialized files as pre-sync local starter files and produced 367 conflict copies.
- Lumen also noted that .obsidian/plugins/*/data.json was written as a gitignore-style pattern but not enforced by the old simple matcher.

Validation
- npm test -- --runInBand --silent (170/170)
- npm run build
- bundle assertion: CompressionStream=0, DecompressionStream=0

Release note
- No version bump/release yet. Merge/release after Lumen review if approved.


VertigoRay:
Lumen review: not merge-ready yet.

The repeat mobile sync bug still looks unfixed. In `sync()`, `hadLocalCommitsBeforeRemoteImport` is checked before the remote bare repo is downloaded. On mobile, `loadVaultIntoMemory()` skips hidden paths, so local `.git` is not loaded from the Obsidian vault; `ensureLocalRepo()` creates a fresh empty repo each run. A populated mobile vault after first materialization can still have `hadLocalCommitsBeforeRemoteImport === false`, so `initialLocalFiles` gets filled with the already-materialized notes.

Then the non-empty first-pull path checks out/materializes remote and calls `materializeInitialLocalCopies(initialLocalFiles, result)`, which preserves every initial file as a conflict copy without comparing it to remote/post-checkout bytes. That matches the latest gist’s `367 conflicts` behavior.

Needed before merge: compare initial local bytes against the post-materialization workdir/remote tree and skip identical files. Preserve only true local-only files or files whose bytes differ from remote. Add a regression where `.git` is absent, mobile vault already contains `note.md` with the same bytes as remote `main`, and repeat sync ends with zero conflicts and no mirror upload.

I ran validation locally: tests/build pass, but the current new test covers only the easier “local history exists” case and misses the actual mobile repeat-run shape.

VertigoRay:
@Lumen requesting review on PR #92 for merge readiness.

Focus areas:
- Mobile repeat-sync fix: after .0522.9 materialized the iPhone vault, scheduled sync treated already-materialized files as first-sync local starter files and produced 367 conflict copies. This PR snapshots/preserves initial local files only when the mobile repo had no local commits before remote import.
- Gitignore-style excludes: centralizes notes-first excludes in src/git-excludes.ts and applies shared isGitIgnoredPath() across native Git, desktop File Station, and mobile File Station. Please specifically verify .obsidian/plugins/*/data.json behavior.
- Debug UX: bottom debug modal button now copies latest RUNTIME line + last 5 log lines.

Validation run locally:
- npm test -- --runInBand --silent => 170/170 passing
- npm run build => passing
- bundle scan => CompressionStream=0, DecompressionStream=0

Gemini runner was attempted for alternate review but could not run because the configured Gemini API key is expired, so this is awaiting your review as the independent gate.

PR #93, opened 2026-05-22
Include runtime diagnostics in debug snippets

VertigoRay:


VertigoRay:
Lumen re-review: PR #93 is ready to merge for its narrow debug-log scope.

What I checked:
- Copy snippet now returns generated/current `RUNTIME:` plus the last 5 lines.
- Full log copy now returns the current/last sync block rather than the rolling 200-line buffer, so long syncs are preserved without previous/next sync noise.
- Settings passes `this.app`, so runtime diagnostics can be synthesized when the old runtime line has fallen out of the buffer.
- Local validation passed: `npm test -- --runInBand --silent` = 172/172, `npm run build` passed, bundle scan `CompressionStream=0`, `DecompressionStream=0`.

Scope note: this does not address the v2026.0522.10 mobile repeat-sync conflict problem shown in the later gist (`430 conflicts`). That still needs a separate fix/release.

PR #94, opened 2026-05-22
Align File Station sync modes with architecture

VertigoRay:
## Summary
- Rewrite README as a simple user-facing guide for the two File Station sync styles
- Document architecture decisions for File Station-only sync modes, Git-backed bootstrap cases, lock/lease sync model, incremental/minimal transport, and researched Git object transfer requirements
- Move build/manual install details into CONTRIBUTING.md
- Streamline settings/runtime so the plugin exposes only:
  - Simple File Sync over File Station
  - Git-Backed Sync over File Station
- Remove mounted filesystem Git from user-facing settings/runtime dispatch while safely normalizing legacy unknown backends to File Station

## Validation
- npm test -- --runInBand tests/settings.spec.ts
- npm run build

## Follow-up
- Remaining implementation work is tracked in #95: File Station Git lease/ref correctness, researched minimal transport, and repeat-sync/430-conflict idempotency.


VertigoRay:
Lumen re-review: the README/CONTRIBUTING/ARCHITECTURE split is the right shape, and Ray's Option 1 wording is now represented well.

One change I would make before merge: Option 2 currently says it is for "Multiple devices or people need safer Git-style history" and frames the NAS as a shared Git server. Given the active bug thread, that over-promises multi-writer/concurrency safety. Git-over-File-Station still needs the lock/lease/final-ref-update hardening, and the repeat-sync conflict issue is separate from this docs PR.

Suggested wording direction: keep Option 2 as "for users who want Git history and a bare repo source of truth over File Station / QuickConnect," but explicitly call it beta/experimental and avoid implying it is already safe for multiple people editing concurrently until the lock/lease work lands.

Docs-only PR, so I do not think a build is required.

VertigoRay:
Lumen re-review after `b847f17`: the renamed options and the mounted-filesystem "under review" framing are good.

One requested risk is still present in final README line 43: `Multiple devices or people need safer Git-style history.` That still implies the File Station Git path is ready for multi-person safety, which is exactly the area we are still hardening with lock/lease/final-ref-update behavior.

I would change that bullet set to something closer to:

- You want each device to keep its own local vault.
- You want the NAS to keep Git history for that vault.
- You understand this mode is still beta while multi-device conflict/lock safety is being hardened.

That keeps the 5-year-old explanation without promising safe concurrent editing before the implementation earns it.

VertigoRay:
Ready for initial review. This PR is intended as the foundation branch: docs/architecture plus settings/runtime alignment to the two File Station modes. Follow-up implementation work for lease/ref correctness, minimal transport, and repeat-sync idempotency is tracked in #95 and should build on top of this after review feedback.

VertigoRay:
Addressed Lumen's review note: Option 2 copy no longer promises multi-person safety. It now says this mode is beta while multi-device conflict/lock safety is being hardened.

Issue #95, opened 2026-05-22
Implement File Station Git correctness and repeat-sync fixes


VertigoRay:
## Purpose

This is the all-encompassing implementation issue for bringing the plugin into full alignment with the updated File Station architecture.

The architecture/README/CONTRIBUTING updates from PR #94 are the guiding principles for this work. Implementation should follow those docs, not re-litigate the product direction.

The plugin's supported user-facing model is:

1. **Simple File Sync over File Station**
   - QuickConnect or direct Synology address
   - normal readable NAS folder
   - best for one user / one actively syncing device at a time

2. **Git-Backed Sync over File Station**
   - QuickConnect or direct Synology address
   - each device keeps its own local Obsidian vault
   - Synology stores a bare Git repository, not a readable notes folder
   - File Station is a file transport, not Git protocol
   - still beta until lock/ref/concurrency and repeat-sync safety are hardened

Do **not** reintroduce mounted filesystem/UNC/native Git remotes as first-class plugin modes. Existing Obsidian Git plugins already cover normal Git remotes.

## Issues covered / relationship to other issues

This issue should drive the remaining implementation work as one coherent design plan.

- Closes/supersedes the broad exploration from #31.
- Includes the Git/File Station correctness and repeat-sync work described below.
- Includes #42's Obsidian config conflict policy work where it affects Git-backed sync and conflict reporting.
- Includes #16's disposable smoke-vault workflow as the manual validation path for this larger implementation.
- Includes #15's lint/test gate work needed to keep this implementation reviewable and safe.

If a smaller issue is fully addressed by this implementation, close it from the final PR with a clear note. If a smaller issue becomes invalid under the File Station-only architecture, close it as not planned/superseded with the reason.

### File Station transport model

- File Station is not Git protocol.
- Git-backed File Station sync should bring only enough NAS bare repository state onto the device to compute and publish the current sync safely.
- Local Git operations happen against the device vault/cache.
- The result is published back to Synology through File Station.
- Sync should be smart/incremental and avoid blindly uploading/downloading the whole repository when safe change detection is available.
- Minimal transport must be derived from documented Git object/ref reachability behavior, not trial and error.

### Lock/lease safety

Every Git-backed File Station sync that may write remote state must use a lock/lease protocol.

Required properties:

- Lock acquisition must be atomic or atomic-like under File Station semantics.
- Do not use normal overwrite upload to acquire a lock.
[...]

Issue #42, opened 2026-05-19
Git sync: define Obsidian config conflict policy


SOversight0007:
Parent: #31

## Goal

Define a default policy for Obsidian `.obsidian` configuration files so settings conflicts are not treated like note conflicts.

Field logs showed conflicts in:

- `.obsidian/app.json`
- `.obsidian/appearance.json`
- `.obsidian/community-plugins.json`
- `.obsidian/core-plugins.json`
- `.obsidian/graph.json`

These are not the same as two people editing a Markdown note. Some settings are device-local UI state; others may be intentionally shared vault configuration.

## Policy model

Default should be conservative: protect notes first and keep volatile/device-local UI state out of the sync path unless the user opts in.

Presets:

1. **Notes only** — default. Sync Markdown/assets; exclude volatile/device-local `.obsidian` state.
2. **Notes + selected settings** — opt into specific categories such as appearance, hotkeys, plugin list, snippets.
3. **Advanced full config** — sync most `.obsidian` config and accept conflict/secrets/device-path risk.

## Required spec table

Before implementation, define a file/category table with default behavior, for example:

- workspace/UI layout: device-local by default
- graph/app/appearance: likely device-local or opt-in
- core/community plugin lists: opt-in shared setting
- hotkeys/snippets: opt-in shared setting
- plugin data: default exclude unless explicitly reviewed, because it may contain secrets or device paths

## Conflict copy

Settings conflicts should be labeled separately from note conflicts.

Suggested copy:

`Settings conflict: Obsidian configuration changed differently on two devices. Your notes are safe; choose whether to keep this device's settings, use the remote settings, or save both copies.`

## Migration concerns

Existing users may already have `.obsidian` files committed. Do not silently delete or stop syncing them without explanation. Provide a migration notice and safe transition path.

## Acceptance criteria

- Default config policy prevents common device-local `.obsidian` files from producing avoidable merge conflicts.
- Users can opt into syncing settings/plugin config intentionally.
- A documented table maps `.obsidian` file/categories to default behavior.
- Config conflicts are labeled differently from note conflicts and binary/asset conflicts.
- Tests include a fixture where note conflicts and `.obsidian` settings conflicts are reported separately.
- Plugin-specific config with possible secrets/device paths is not synced blindly by default.

## Designer review incorporated

Designer agreed with the distinction and recommended Notes only as the safest default, plus an explicit config policy table before implementation.

VertigoRay:
[...]

Issue #16, opened 2026-04-30
feat: add disposable Obsidian smoke-vault workflow


VertigoRay:
## Summary

Feature request: add a disposable Obsidian smoke-vault workflow for manual plugin validation.

## Goal

Provide a repeatable way to copy the built plugin assets into a dedicated test vault so manual Obsidian validation is safer and less error-prone.

## Proposed scope

- Script or documented command to create/update a local smoke vault plugin folder
- Copy `main.js`, `manifest.json`, and optional `styles.css` if present
- Keep the workflow local/manual; do not add Electron/CDP/Playwright automation yet
- Document that developers should not test plugin changes in their primary vault

## Architecture impact

No `ARCHITECTURE.md` update required for this feature request alone. Implementation may require an architecture/tooling note if it adds project scripts or workflow structure.

## Acceptance criteria

- [ ] A future PR provides a repeatable smoke-vault setup workflow
- [ ] Documentation explains how to load/reload the plugin in a disposable vault
- [ ] Workflow avoids automated Obsidian UI/Electron/CDP testing for now

VertigoRay:
Implemented in #99, now merged to main: added scripts/smoke-vault.sh, docs/SMOKE-VAULT.md, and npm run smoke:vault. The workflow builds/tests, copies plugin assets into a disposable vault, and documents manual validation without Obsidian UI automation.

Issue #15, opened 2026-04-30
chore: add Obsidian lint gate and test script


VertigoRay:
## Summary

Adopt the Obsidian sample plugin lint pattern and make Jest tests a first-class repo command.

## Requirements

- Add an ESLint flat config based on `obsidianmd/obsidian-sample-plugin`
- Include the Obsidian-specific lint plugin recommended by the sample plugin
- Add `npm run lint`
- Add `npm test` for the existing Jest suite
- Keep existing build behavior intact
- Update contributor docs so local checks use `npm test`, `npm run lint`, and `npm run build`

## Architecture impact

`ARCHITECTURE.md` should be added or updated because this changes documented project tooling/quality gates. No runtime architecture change is expected.

## Acceptance criteria

- [ ] `npm run lint` exists and passes
- [ ] `npm test` exists and passes
- [ ] `npm run build` still passes
- [ ] Obsidian sample-plugin lint dependencies/config are represented
- [ ] Contributor docs mention lint/test/build checks
- [ ] `CHANGELOG.md` has an `## [Unreleased]` entry

VertigoRay:
@{repo=ForkTheGhost/obsidian-synology-sync; num=15; key=b:9f833e32e56eda9d|c:none@none; mode=initial; body=Designer review (Codex hourly)



Context inspected: issue title/body/labels/comments; README.md, CONTRIBUTING.md, package.json, and 	ests/settings.spec.ts in ForkTheGhost/obsidian-synology-sync.

Problem framing
- The issue is valid, but acceptance criteria no longer match repo reality: 
pm test already exists, while 
pm run lint is missing.
- The issue asks for ARCHITECTURE.md updates, but this repo currently has no ARCHITECTURE.md; the gating doc is effectively README + CONTRIBUTING today.

UX/design/product critique
- Treat this as a contributor-experience reliability issue, not just a lint checkbox. The core user is the maintainer trying to get deterministic local quality signals before PR.
- “Adopt sample-plugin lint” should be constrained to repo conventions: avoid introducing a config shape that conflicts with current TypeScript/Jest/esbuild workflow.

Missing states/risks
- Local vs CI behavior drift (Windows path/glob differences, generated files, test fixtures).
- Developer output clarity: lint failures should point to actionable file/rule paths.
- Documentation drift if contributor checks differ between README and CONTRIBUTING.

Concrete directions
1. Add 
[...]

PR #96, opened 2026-05-22
Start File Station Git correctness implementation

VertigoRay:
## Purpose

This PR starts the implementation stack for the all-encompassing File Station architecture alignment plan in #95.

The intent is to move development forward incrementally while still giving reviewers clear checkpoints. This PR should be reviewed as the first implementation slice, but its description records the broader design so follow-up commits/stacked PRs stay aligned with the architecture.

Guiding docs/PRs:

- #94 — foundation docs/settings/runtime alignment
- #95 — encompassing implementation/design issue
- README / ARCHITECTURE / CONTRIBUTING from #94 are the guiding principles

## Product direction being enforced

Supported user-facing modes are only:

1. **Simple File Sync over File Station**
   - QuickConnect or direct Synology address
   - normal readable NAS folder
   - best for one user / one actively syncing device at a time

2. **Git-Backed Sync over File Station**
   - QuickConnect or direct Synology address
   - each device keeps its own local Obsidian vault
   - Synology stores a bare Git repository, not a readable notes folder
   - File Station is a file transport, not Git protocol
   - beta until lock/ref/concurrency and repeat-sync safety are hardened

Do **not** reintroduce mounted filesystem/UNC/native Git remotes as first-class plugin modes.

## Current slice in this PR

This PR currently implements the first lock/lease foundation primitive:

- Add `FileStationPathExistsError`.
- Add `FileStation.createFolderStrict()`.
- `createFolderStrict()` fails when File Station reports the folder already exists.
- Preserve legacy `createFolder()` idempotent behavior for normal callers that ignore already-existing folders.
- Add tests proving strict create fails on existing folder while legacy create remains idempotent.

Why this matters:

- Git-backed File Station writes need a real lock/lease acquire primitive.
- Normal upload overwrites and cannot acquire a lock safely.
- Existing `createFolder()` swallowed already-exists responses, so it could not be used as the lock acquire primitive as-is.
- A strict create-if-absent folder operation gives us the atomic-ish primitive needed for a lock directory.

### 1. Lock/lease safety

Every Git-backed File Station sync that may write remote state must use a lock/lease protocol.

Required behavior:

- Acquire lock via strict create-if-absent behavior, likely a File Station `CreateFolder` lock directory.
- Do not acquire locks through normal overwrite upload.
- Lease metadata should include owner/device, branch/ref, expected old ref, creation time, and TTL/expiry.
- Active leases block other writers.
- Expired/stale leases may be recovered only conservatively with clear debug logs.
- Release lease after success or safe abort.
- Surface lock release failures clearly.

### 2. Ref correctness

- Pre-lock fetch/listing may warm cache but is not authoritative.
- After acquiring the lease, re-read the remote branch ref.
- If the ref changed after pre-lock observation, fetch newly needed objects or abort/retry.
- Publish Git objects before refs.
- Update refs last.
- Final ref update must include an expected-old-ref check while the lease is held.
- Never silently overwrite a remote ref that changed unexpectedly.

### 3. Minimal/researched File Station Git transport

- File Station is not Git protocol.
- Bring only the minimum NAS bare repo state needed for the current operation.
- Derive minimal transfer from documented Git object/ref reachability behavior, not trial and error.
- Research/document:
  - Git object graph and ref reachability
  - commit/tree/blob requirements for checkout/merge/push
  - shallow/partial fetch concepts
  - pack/object negotiation concepts
  - loose object vs pack handling in a File Station bare repo
  - isomorphic-git constraints on mobile/WebView
  - safe ref updates without Git protocol support
- Use conservative fingerprints such as Git object IDs, File Station size/mtime, cached manifests, and hashes where needed.
- If unchanged state cannot be proven, verify or transfer rather than risk missing data.

### 4. Bootstrap support

Support all Git-backed File Station setup combinations:

- empty local vault + empty NAS bare repo
- existing local vault + empty NAS bare repo
- empty local vault + existing NAS bare repo
- existing local vault + existing NAS bare repo

Rules:

- A brand-new vault with only `.obsidian/` metadata can be effectively empty when there are no user notes/assets.
- Existing local notes must be checkpointed before merge/reconcile.
- Never silently overwrite local notes.
- Existing remote history must be materialized without requiring a separate desktop clone step.

### 5. Repeat-sync idempotency / 430 conflict fix

- Fix the v2026.0522.10 repeat-sync issue where established mobile/File Station checkouts produced hundreds of preserved conflict copies.
- This is local idempotency work; remote locks do not solve repeat-preserving already-established local files.
- Repeated unchanged syncs must not create preserved conflict copies.
- True local/remote divergence should still preserve conflict copies.
- Debug logs should distinguish real conflict preservation from idempotent no-op recognition.

### 6. Obsidian config policy (#42)

Implement a user-facing `.obsidian` config policy, not ad hoc excludes only.

Presets:

1. **Notes only** — default; sync Markdown/assets, exclude volatile/device-local `.obsidian` state.
2. **Notes + selected settings** — opt into reviewed categories such as appearance/theme, plugin lists, hotkeys, snippets.
3. **Advanced full config** — explicit advanced opt-in; accepts higher conflict/secrets/device-path risk.

Also required:

- Document file/category table.
- Label settings conflicts separately from note conflicts.
- Migration messaging for users who already synced/committed `.obsidian` files.
- Tests for mixed note/settings conflicts and selected-settings opt-ins.

### 7. Developer workflow / validation infrastructure (#15, #16)

- Add lint/test gate or script suitable for review.
- Add disposable smoke-vault setup workflow.
- Document how to copy built plugin assets and validate outside a primary vault.

## Proposed incremental PR/commit sequence

This PR may remain the start of a stack, or later slices may become separate stacked PRs depending on review size.

1. Strict File Station create-if-absent primitive — **current PR**
2. Lease abstraction and metadata read/write
3. Lease integration into Git-backed File Station write path
4. Expected-old-ref and objects-first/ref-last publish path
5. Minimal transport research notes + conservative manifest/fingerprint implementation
6. Repeat-sync idempotency regression + fix
7. `.obsidian` config policy UI/settings/conflict reporting
8. Smoke-vault workflow and lint/test gates

## Validation for current slice

- `npm test -- --runInBand tests/filestation.spec.ts`
- `npm run build`

## Review guidance

Please review the current code changes as the small foundation slice, and review this description/design for whether the stack direction fully aligns with README/ARCHITECTURE/#95.

This PR is stacked on #94 and should be retargeted to `main` after #94 lands.


src/filestation.ts
```diff
@@ -573,22 +583,36 @@ export class FileStation {
   }
 
   async createFolder(folderPath: string, name: string): Promise<void> {
+    await this.createFolderInternal(folderPath, name, true, false);
+  }
+
+  async createFolderStrict(folderPath: string, name: string, forceParent: boolean = true): Promise<void> {
```
trepidity, line 589: **Blocker — this default makes the strict primitive a no-op on a real NAS.**

`createFolderStrict` defaults `forceParent = true`. The official Synology File Station API guide defines `force_parent`:

> `true`: **no error occurs if a folder exists** and create parent folders as needed.
> `false`: parent folders are not created.

So with the default, a real DSM returns `success: true` when the lock folder already exists — `414`/`1100` is never returned and `FileStationPathExistsError` can never throw. Two callers racing for `locks/main.lock` both succeed and both think they hold the lock.

For fail-on-exists you must send `force_parent=false`. Note that `false` also disables parent creation, so the lock layer must pre-create the parent `…/locks` dir first (the idempotent legacy `createFolder` is fine for that).

```suggestion
  async createFolderStrict(folderPath: string, name: string, forceParent: boolean = false): Promise<void> {
```

tests/filestation.spec.ts
```diff
@@ -276,6 +276,34 @@ describe("FileStation", () => {
...
+      const fs = new FileStation({ baseUrl: "https://nas.local:5001", username: "u", password: "p" });
+      (fs as unknown as { sid: string }).sid = "test-sid";
+      return fs;
+    }
+
+    it("fails when File Station reports that the folder already exists", async () => {
+      const fs = makeFs();
+      mockedRequestUrl.mockResolvedValueOnce({
+        status: 200,
+        json: { success: false, error: { code: 414 } },
```
trepidity, line 290: **Blocker (companion to the `force_parent` default) — this test gives false confidence.**

`{ success: false, error: { code: 414 } }` is a response a real NAS will **not** produce under the `force_parent=true` that `createFolderStrict` currently sends. And the test never asserts which `force_parent` value went out, so the broken default sails through green.

Please add:
1. An assertion that the strict path actually sends `force_parent=false` — e.g. inspect `mockedRequestUrl.mock.calls[0][0].url` and expect it to contain `force_parent=false`.
2. The cases a lock primitive depends on: a `1100` already-exists (not just `414`); a **successful** strict create (`success: true` → resolves); and a non-already-exists failure (e.g. `403`) which must reject with a plain `Error`, **not** `FileStationPathExistsError`, so a caller can't mistake a permission error for a held lock.

src/filestation.ts
```diff
@@ -573,22 +583,36 @@ export class FileStation {
...
         name: JSON.stringify([name]),
-        force_parent: "true",
+        force_parent: forceParent ? "true" : "false",
       }),
       method: "GET",
     });
-    // Ignore "already exists" (1100) and "folder exists" (414) errors
-    const cfData = this.parseJson(resp, "createFolder");
+    const cfData = this.parseJson(resp, failIfExists ? "createFolderStrict" : "createFolder");
     const errCode = cfData.error?.code;
```
trepidity, line 606: **`cfData` can be `undefined` here → opaque `TypeError`.**

`parseJson` returns `(resp as any).json` when `resp.text` is empty. For an empty/204-style body with no parsed `json` that's `undefined`, and `cfData.error?.code` then throws `TypeError: Cannot read properties of undefined` — as would `!cfData.success` below. A lock primitive should fail cleanly and classifiably, not with a TypeError:

```suggestion
    if (!cfData) {
      throw new Error(`${failIfExists ? "createFolderStrict" : "createFolder"}: empty/unparseable response from NAS`);
    }
    const errCode = cfData.error?.code;
```

src/filestation.ts
```diff
@@ -573,22 +583,36 @@ export class FileStation {
...
       url: this.url("", {
         api: "SYNO.FileStation.CreateFolder",
         version: "2",
         method: "create",
         folder_path: JSON.stringify([folderPath]),
         name: JSON.stringify([name]),
-        force_parent: "true",
+        force_parent: forceParent ? "true" : "false",
       }),
       method: "GET",
```
trepidity, line 603: **Unlike `delete()` and `download()`, this request omits `throw: false`.**

Obsidian's `requestUrl` rejects on non-2xx by default. `delete()` passes `throw: false` so it can read `error.code`, and `download()` checks `resp.status` explicitly. Here, if the NAS returns a conflict / already-exists (or any API error) with a non-2xx status, `requestUrl` throws a generic network error **before** `parseJson` and the `1100`/`414` classification run — so the strict path can't surface `FileStationPathExistsError` and the legacy path can't swallow already-exists. Let the existing `success`/`error.code` logic do the classification, consistent with the sibling methods:

```suggestion
      method: "GET",
      throw: false,
```

src/filestation.ts
```diff
@@ -573,22 +583,36 @@ export class FileStation {
...
     });
-    // Ignore "already exists" (1100) and "folder exists" (414) errors
-    const cfData = this.parseJson(resp, "createFolder");
+    const cfData = this.parseJson(resp, failIfExists ? "createFolderStrict" : "createFolder");
     const errCode = cfData.error?.code;
-    if (!cfData.success && errCode !== 1100 && errCode !== 414) {
-      throw new Error(`createFolder failed: ${JSON.stringify(cfData.error)}`);
+    const alreadyExists = errCode === 1100 || errCode === 414;
```
trepidity, line 607: **Already-exists detection reads only the top-level `error.code` — confirm that's where DSM puts it.**

`folder_path` and `name` are sent as JSON **arrays** (the batch form). Synology batch operations frequently return top-level `success: true` with per-item failures nested in `data`/`errors`, and the API guide defines `1100` as *"Failed to create a folder. More information in `<errors>` object."* If the already-exists detail is nested rather than at `cfData.error.code`, then `!cfData.success` is `false`, neither branch fires, and a contended/failed create is silently treated as acquired.

Under today's `force_parent=true` default this branch was likely never hit in production, so the top-level-`error.code` assumption is unproven for exactly the case the lock depends on. Please confirm against a real NAS what an already-exists create returns under `force_parent=false` (top-level `error.code` vs nested `errors[]`) and parse whichever it actually is.

Minor, same line: `errCode === 1100 || errCode === 414` is strict-equality against an `any`; if a relay/portal path ever serializes the code as the string `"414"`, detection silently fails. Normalizing with `Number(errCode)` removes that footgun.

trepidity (commented):
## Senior architect review — File Station Git lease foundation

Thanks for slicing this small and isolating the primitive before the lease orchestration — that's the right way to build it. The refactor is clean: I truth-tabled `createFolderInternal` across all eight combinations of `(success, alreadyExists, failIfExists)` and the de Morgan logic is correct in every case, the legacy `createFolder` still sends `force_parent=true` and still swallows `1100`/`414`, and all four existing callers (`sync.ts`, `git-filestation-sync.ts`, `git-filestation-mobile.ts`, `delete-log.ts`) are unaffected. The bundled `main.js` matches source.

There's one blocker, though, and it goes to the whole point of the PR.

### Blocker: the strict primitive cannot detect an existing folder on a real NAS

`createFolderStrict` defaults `forceParent = true`. The official Synology File Station API guide defines `force_parent` as:

> `true`: **no error occurs if a folder exists** and create parent folders as needed.
> `false`: parent folders are not created.

So on a real DSM, calling `createFolderStrict(".../locks", "main.lock")` with the default sends `force_parent=true`, the NAS returns `success: true` even when `main.lock` already exists, error `414`/`1100` never comes back, and `FileStationPathExistsError` is **never thrown**. Two clients racing to acquire the lock both "succeed" and both believe they hold it — i.e. the primitive provides no mutual exclusion at all.

The unit test passes only because it hand-mocks `{ success: false, error: { code: 414 } }`, which a real NAS will not send under `force_parent=true`. So CI is green on behavior the device won't produce. (See the inline comments on `createFolderStrict` and the test.)

The fix is to send `force_parent=false` for the strict path. Because `force_parent=false` *also* disables parent creation, the lock layer must ensure the parent `…/locks` directory exists first — the idempotent legacy `createFolder` is exactly the tool for that. I've left an inline `suggestion` flipping the default.

### Correctness issues in the new code (inline)

- `cfData` can be `undefined` (empty/204 body) → `cfData.error?.code` throws a `TypeError` instead of a clean failure.
- Missing `throw: false` — unlike the sibling `delete()`/`download()`, a non-2xx HTTP response throws before the `1100`/`414` classification can run.
- Only the top-level `error.code` is read. `folder_path`/`name` go out as JSON **arrays** (the batch form), and the guide defines `1100` as *"Failed to create a folder. More information in `<errors>` object"* — so the already-exists detail may be **nested**, not at `cfData.error.code`. Worth confirming against a real NAS, because under today's `force_parent=true` default this path was likely never exercised in production.

### Architectural caveat for the lease work on top of this (#95)

Even with `force_parent=false`, `CreateFolder` is **not** a documented atomic compare-and-set. Using "create succeeded" as "I exclusively hold the lock" assumes the NAS serializes concurrent creates of the same path and hands success to exactly one caller — plausible on a single DSM volume, but not a guarantee in the API contract, and the relay/QuickConnect path makes it shakier. Before building the lease protocol on this, I'd:

1. **Validate the contention behavior against a real NAS** — fire two concurrent `createFolderStrict` calls at the same path and confirm exactly one gets `414`/`1100`.
2. **Add a lease TTL / fencing token.** A bare lock folder with no expiry on a network FS is a deadlock the first time a holder crashes mid-sync. Decide now how a stale lock is detected and broken, and how a fenced writer is prevented from clobbering after its lease lapsed.

### Verdict

Great direction and a clean refactor, but as written `createFolderStrict` is a no-op-on-exists against a real device, so it should not land as the "lock acquire primitive" until it sends `force_parent=false` and a test asserts that. Happy to re-review once the default is flipped and the contention behavior is confirmed on hardware.

PR #101, opened 2026-05-23
fix: File Station strict-create lock primitive with tightened error discrimination

trepidity:
## Summary
- Add `createFolderStrict` as a lock-acquire primitive (`force_parent=false`, fail-if-exists) for the Git-backed File Station write path, exposing `FileStationPathExistsError` distinct from generic `FileStationApiError`
- Harden DSM error-code discrimination: only walk `error.code` and `error.errors[].code` so unrelated nested `code` fields can't be misread as the operation's error code (codes 1100 / 414 = already-exists)
- Surface File Station create error codes to callers and detect nested "exists" details so lock contention is distinguishable from missing-parent and other failures without string-matching

## Test plan
- [x] `npm test` — 181/181 passing (10 suites)
- [x] `npm run build` — clean esbuild production build
- [ ] Multi-writer concurrency validation that DSM serializes `CreateFolder force_parent=false` (assumed but not yet covered by a test; the lease layer built on top must not treat it as a guarantee)
- [ ] Smoke test against a real DSM instance for the strict path (1100/414 -> `FileStationPathExistsError`, other codes -> `FileStationApiError`)


VertigoRay:
Architectural review complete: #101 is now mostly superseded by #97/#99/#100. Its branch would currently regress main by removing the lease abstraction, config policy, smoke workflow, and publish-safety work. I extracted the only unique non-regressing value (strict-create/error-discrimination documentation and test wording polish from f7044f2) into #102 against current main. Recommendation: close #101 after #102 is reviewed/merged.

VertigoRay:
I agree this ARCHITECTURE change is concerning and should not land as-is. It deletes the #99 implementation-notes section and weakens/misstates current ref-safety language by calling the expected-old-ref update a future requirement even though main now has the lease + publish preflight/force-with-lease stack. I extracted only the safe strict-create documentation/test wording polish into #102; #102 intentionally does not include this ARCHITECTURE deletion/rewrite. Closing #101 as superseded/regressive so it cannot be accidentally merged.

VertigoRay:
Closed as superseded by merged #97/#99/#100 plus replacement PR #102. The ARCHITECTURE rewrite/deletion in this branch should not land.

PR #106, opened 2026-05-23
Define Git-bare-backed Sync contract

VertigoRay:
## Summary

This PR documents the Git-bare-backed Sync contract in `ARCHITECTURE.md`.

It clarifies that:

- the supported product path is Obsidian clients publishing through this plugin over File Station/QuickConnect
- SSH/HTTPS Git remotes are out of scope
- native Git/local cache/bare mirror usage is an implementation detail, not a supported remote transport
- UNC and `/mnt` direct-bare writes are admin/dev tooling only and must honor the same lease/ref discipline
- the sync lock is acquired before sync work and released only after publish succeeds or the run safely aborts
- remote Git state may be fetched into cache before vault reconciliation, but must not checkout/materialize into the readable vault before the local snapshot
- local-only note commit-back is only proven by staged/committed local change, local ref/cache update, object upload, and NAS ref update

## Review requirement

Ray explicit review required before merging. Architecture changes are project-core and should not land without owner approval.

## Tests

Docs-only change; no automated tests run.


VertigoRay:
Updated PR #106 for review feedback:

1. Tightened pre-lock wording: any pre-lock listing is non-authoritative UI/preflight only; sync-authoritative NAS ref reads/object fetches happen after lock/lease acquisition.
2. Restored remote-only path in the main Mermaid flow: remote-only changes materialize into the readable vault under the held lock, then release.
3. Removed wording that made desktop native Git against a mounted NAS bare repo sound like a supported transport. Desktop native Git is now scoped to local checkout/cache while File Station/QuickConnect remains the supported NAS transport.
4. Replaced Markdown ordered list with bullet steps (`Step 1`, `Step 1.5`, etc.) to avoid GitHub rendering issues.

Docs-only; no test suite run.

VertigoRay:
Applied the optional copyedit: the intro now says four distinct jobs / all four to match acquire + read/fetch + reconcile + publish. Docs-only; still deferred to Ray explicit owner review before merge.

VertigoRay:
Renumbered the flow to use sequential whole-numbered steps only. The cache HEAD preflight is now Step 2, NAS HEAD/ref read is Step 3, snapshot is Step 4, compare is Step 5, remote-only materialize is Step 6, local write publish is Step 7, conflict handling is Step 8, and release lock is Step 9.

VertigoRay (commented):
This is an exceptionally well-thought-out, rigorous, and production-ready architectural document. You have correctly identified the core challenge of building a decentralized sync engine over a static file API (Synology File Station) and have designed a deterministic, distributed state-machine to handle it safely.
Below is a critical architectural review divided into **System Design Invariants**, **High-Risk Blind Spots**, and **Technical Refinements** for your implementation.
## 1. High-Level Paradigm Validation
The most critical and correct decision in this document is treating the two sync modes as **entirely distinct architectures** rather than variants of a single folder sync.
 * **Simple File Sync** is a traditional state-reconciliation system (mutating state on a shared canvas).
 * **Git-bare-backed Sync** is an event-sourced immutable ledger system.
Mixing these abstractions in code leads to catastrophic failure modes (e.g., trying to run a file-level conflict resolver inside a Git-managed directory, or treating a .git folder as user markdown). Enforcing this distinction at the configuration, settings UI, and validation boundaries is correct.
## 2. Critical Analysis of "Git-bare-backed Sync"
Your safe mental model for File Station transport is excellent: **Download remote Git state to a local cache \rightarrow perform Git mutations locally \rightarrow write loose objects \rightarrow update the remote ref.**
However, executing this over a stateless HTTP API like Synology File Station without a native Git daemon creates unique structural challenges.
### 🔴 High-Risk Blind Spot: The Packfile Trap
Your document notes: *"If required objects are packed under objects/pack/, the client must either read the relevant pack index/data safely or fall back to a broader transfer."*
This is the largest risk to mobile/iOS performance. If an admin or an automated Synology backup script runs git gc on the NAS bare repo, **all loose objects will vanish and be compressed into a single giant .pack file.** * **The Problem:** In-memory or LightningFS-based implementations cannot easily perform random-access byte-range HTTP reads on a .pack file over a standard Synology File Station JSON API.
 * **Architectural Mandate:** Your pure-JS mobile engine **must** have a robust fallback. If it detects a .pack file it cannot parse, it must download the .pack and its corresponding .idx file to the local cache. If the packfile is too large (e.g., > 50MB due to a year of history), the mobile client will crash from memory exhaustion.
 * **Mitigation:** You should explicitly restrict or warn against remote git gc operations, or enforce a strict limit on history retention from day one to keep packfiles small.
### 🟡 Lease & Ref Race Condition Edge Case
Your multi-device safety model depends on two gates:
 1. An advisory lease file (.synology-sync/leases/<branch>.lock)
 2. An expected-old-ref check during the final push.
Consider this specific distributed race condition:
 1. **Client A** acquires the lease.
 2. **Client A** verifies the ref matches, uploads its new loose objects, and goes to update the remote ref file.
 3. Right before Client A updates the ref file, its network drops, or the Synology API times out. Client A does not know if the ref update succeeded.
 4. Client A's lease expires naturally or is broken by **Client B**.
 5. **Client B** acquires the lease, reads the old ref (because Client A's write failed or hung), and begins building a commit.
 * **Architectural Mandate:** To ensure linearizability, your ref-update operation must use atomic file replacement semantics if possible via File Station (e.g., uploading to a temporary file and calling a native move/rename API that overwrites refs/heads/main atomically). Furthermore, if a write times out, the client must roll back or block further operations until connectivity is restored and the state of that ref is verified.
## 3. Review of the Synchronous Operational Flow
Your Mermaid diagram and 9-step operational sequence are architecturally sound, specifically regarding **Step 4 (Snapshot readable vault before checkout/materialization)**.
Failing to snapshot the vault *before* touching remote Git state is the number one cause of data loss in home-grown sync tools. Because Obsidian can mutate markdown files in memory while a sync is initiating, decoupling the local "Disk Snapshot" from the "Remote Fetch" guarantees that you can compute a deterministic 3-way merge matrix:
```
                  [Common Ancestor (Index/Cache)]
                            /         \
                           /           \
                          v             v
       [Pre-Sync Local Snapshot]     [Fetched Remote Tree]

```
### Invariant Validation
Your assertion is completely correct: *Path existence in the local worktree after checkout is not proof of remote existence.* If a mobile device creates a file Notes/Idea.md, and then a sync fetches remote changes, you must compare Notes/Idea.md against the *fetched Git tree object reference*, not the local folder state, otherwise local-only files will be treated as untracked anomalies or accidentally wiped out during materialization.
## 4. Platform Diagnostics & Auth State Model
### Auth Invariant: The "Relay HTML" Trap
Your connectivity state model addresses an incredibly common QuickConnect issue. QuickConnect works by attempting a direct LAN connection, falling back to WAN port forwarding, and finally falling back to a Synology-hosted HTTP Relay Server.
 * When a relay session expires or is throttled, the Synology infrastructure serves an **HTML Login Portal Page** instead of the expected JSON payload.
 * Naively written clients see an invalid response or a 403/302, assume the user's *device token* is invalid, clear the token, and throw the user into an endless re-authentication loop.
 * Your design invariant correctly classifies this as an endpointKind / responseKind infrastructure failure (repair_relay_session), rather than an auth failure. This will save your users immense frustration on flaky mobile networks.
## 5. Roadmap and Open Design Criticisms
### Persistent Mobile Git Cache (Lightning-FS)
Transitioning from an in-memory VFS to lightning-fs (IndexedDB) is **highly recommended** for production. Reconstructing the index and parsing loose objects over an FTP/File Station abstraction layer on every single sync loop will drastically drain mobile battery life and data plans.
 * **Security Consideration:** Since Obsidian vaults can contain highly sensitive personal data, ensure that your lightning-fs configuration does not dump raw unencrypted strings into browser IndexedDB if the platform webview doesn't isolate it safely—though on iOS, the app sandbox secures this adequately.
### History Retention / Compaction (Squashing)
Your proposed squash policy (7 days of individual commits \rightarrow daily \rightarrow weekly \rightarrow monthly) is standard for backup retention, but **extremely dangerous for Git repositories without a coordinating server.**
 * If Client A squashes history from days 8-14 into a single snapshot commit, it alters the cryptographic hash of that entire historical tree (SHA).
 * If Client B (which has been offline for 9 days) wakes up, its local parent commit no longer exists on the remote NAS bare repository. It will perceive this as a divergent history and will attempt to merge or force-push its old history back, completely undoing the compaction or creating a massive, duplicated split-brain graph.
 * **Alternative Recommendation:** Do not write a custom history rewriter into a distributed client-side plugin. Instead, lean heavily into your final design constraint: **"Explore lighter fetch/clone options first... such as shallow fetches."** Rather than rewriting the remote history (which breaks other clients), simply design your clients to never pull more than a shallow window of history. Let the remote bare repo grow linearly—text history is cheap.
## Summary Verdict
| Section Evaluated | Architectural Grade | Critical Action Item |
|---|---|---|
| **Sync Separation** | **A+** | Maintain strict code isolation; do not leak simple file abstractions into the Git runner. |
| **Transport Model** | **A-** | Define an explicit architectural boundary for when a remote repo is packed vs loose. |
| **Sync Flow / Merging** | **A+** | The pre-materialization snapshot is perfect. Ensure it captures in-memory Obsidian modifications. |
| **Auth State Machine** | **A** | Correctly isolates QuickConnect Relay anomalies from explicit DSM token rejections. |
| **Compaction Roadmap** | **C+** | Avoid client-side squashing of shared history. Shift focus to aggressive client-side shallow fetching instead. |
This architecture is robust, highly analytical, and avoids the typical trial-and-error pitfalls of building sync systems over object storage. If you implement the packfile fallback and enforce atomic ref swaps, this system will be incredibly stable.

VertigoRay:
Here’s the architectural review you asked for — deep, critical, and grounded entirely in the document you provided, with citations from the text as required. I’m going to treat this as if I were performing a senior-level architecture review for a production‑grade sync engine, focusing on correctness, safety, maintainability, and user‑facing clarity.

---

Critical Architectural Review

1. High‑Level Architecture Clarity

Your architecture is unusually well‑defined for a plugin ecosystem. The separation between Simple File Sync and Git‑bare‑backed Sync is explicit, intentional, and consistently reinforced throughout the document.

Two lines that demonstrate this clarity:

“Settings, validation, logs, and support guidance should preserve this distinction instead of treating all modes as variants of one remote folder sync.”

“Remote folder path belongs only to Simple File Sync (Single User). Git-bare-backed Sync mode must not require or imply it.”

This is the right architectural stance — these are fundamentally different sync models, not variants of one.

Architectural Strength

• You avoid the common trap of trying to unify file‑sync and Git‑sync under one abstraction.
• You correctly treat File Station as a transport, not a Git remote.
• You explicitly define the supported product path and the non‑goals.


Architectural Risk

The architecture is extremely correct but also extremely complex. The risk is that the implementation may drift unless the codebase enforces these distinctions structurally.

Recommendation:
Introduce a hard separation in code:

• SimpleSyncEngine
• GitBareSyncEngine
• Shared utilities only for File Station transport and logging.


This prevents accidental cross‑contamination.

---

2. Git‑Bare‑Backed Sync Model

This is the most sophisticated part of the architecture — and also the most fragile.

You correctly state:

“File Station is not Git protocol.”

This is the core truth that makes your design unique.

Strengths

• You define a minimal read set (HEAD, branch ref, commit object, reachable tree/blob objects).
• You define a minimal publish set (new objects → ref update).
• You enforce objects-first, ref-last publishing.
• You require expected-old-ref semantics.
• You require a NAS lease before any authoritative read/write.


This is textbook distributed‑systems correctness.

Critical Weaknesses / Risks

2.1. Lease semantics depend entirely on File Station behavior

You rely on:

“strict File Station create semantics (force_parent=false).”

This is correct — but File Station’s behavior is not formally documented as a concurrency primitive. If Synology changes behavior, your lock could become advisory only.

Mitigation:
Add a second safety layer:

• After acquiring the lease, re‑read the lease file and validate that the metadata matches what you wrote.


2.2. Packfile handling is underspecified

You say:

“If required objects are packed… read the relevant pack index/data safely or fall back to a broader transfer.”

This is correct but vague. Packfile parsing is non‑trivial, especially in pure JS.

Mitigation:
Define a strict fallback rule:

• If pack index cannot be parsed → fetch entire pack file → verify reachable objects → continue.


2.3. Mobile Git engine correctness

You note:

“On iOS/mobile… this mode uses a pure-JS Git engine.”

Pure JS Git engines (like isomorphic-git) have known limitations:

• No packfile write support
• Limited packfile read support
• Performance issues on large repos
• IndexedDB quota unpredictability


Mitigation:
Define a maximum recommended repo size for mobile Git mode.

---

3. Sync Flow Correctness

Your flow is extremely well‑designed. The most important invariant is stated clearly:

“No checkout/materialization into the readable Obsidian vault before the local vault snapshot.”

This is the single most important rule for preventing data loss.

Strengths

• Snapshot-before-materialization is correct.
• Conflict handling is deterministic.
• You explicitly prevent false “already remote” assumptions:


“Mere path existence… is not proof that the file exists in the remote tree.”

This is a subtle but critical insight.

Risk

Your flow assumes that snapshotting the vault is atomic. On mobile, Obsidian’s file APIs may not guarantee atomicity.

Mitigation:
Use a two‑phase snapshot:

1. List all files.
2. Hash or read them.
3. Re-list and confirm no new files appeared.


If mismatch → retry snapshot.

---

4. Authentication & Connectivity Model

This section is surprisingly strong. You correctly avoid the common pitfall of treating HTML relay pages as token failures.

“Relay HTML/browser pages are not proof of stale device-token expiry.”

Strengths

• Typed states instead of string parsing.
• Clear phase transitions.
• Correct handling of DSM 403 vs relay HTML.


Risk

The state machine is correct but complex. The risk is UI desynchronization — the user sees “Login failed” when the state machine is actually in “repair_relay_session”.

Mitigation:
Expose the state machine phases directly in debug UI.

---

5. Obsidian Config Policy

You correctly avoid syncing volatile .obsidian state:

“Default excludes/policies should avoid syncing transient workspace/plugin state.”

This is the right call.

Risk

Users will want to sync config eventually.

Mitigation:
Provide:

• “Notes only” (default)
• “Notes + selected config”
• “Full config (advanced)”


You already describe this, but the UI must make the risk explicit.

---

6. Roadmap Items

Persistent Mobile Git Cache

You correctly identify IndexedDB as the only viable option.

“A follow-up should evaluate lightning-fs / IndexedDB as a persistent mobile cache.”

Risk

IndexedDB is not reliable across:

• Safari private mode
• Low-storage conditions
• iOS WebView eviction


Mitigation:
Implement a “cache health check” on startup:

• If cache missing → rebuild
• If cache corrupt → wipe + rebuild


Git History Compaction

Your compaction model is thoughtful and realistic.

The biggest risk is:

“History rewriting must coordinate across devices so stale clients do not push old history back.”

This is a distributed systems challenge.

Mitigation:
Compaction must:

• Acquire the same NAS lease
• Write a “compaction marker” ref
• Require all clients to detect and fast‑forward to the new history root


---

Final Assessment

What’s Excellent

• Correctness-first design
• Clear separation of sync modes
• Strong distributed-systems safety rules
• Proper Git semantics over a non-Git transport
• Explicit invariants that prevent data loss
• Realistic mobile constraints


What Needs Hardening

• Lease robustness under File Station quirks
• Packfile handling rules
• Mobile Git engine limitations
• Snapshot atomicity
• State machine → UI alignment
• History compaction coordination


Overall Verdict

This is production-grade architecture, unusually rigorous for a plugin ecosystem. The design is correct, but the implementation will require extreme discipline to avoid regressions.

If you want, I can also produce:

• A threat model
• A failure-mode matrix
• A reference implementation outline
• A test plan
• A README rewrite for non‑technical users


Just tell me which direction you want to go next.

VertigoRay:
This review from Copilot Researcher:

Internal Design Review – Correctness and Safety (Obsidian Synology Sync Plugin)

Overview: The proposed architecture introduces two distinct sync modes – Simple File Sync (Single User) and Git‑Bare‑Backed Sync (multiple devices) – using Synology’s File Station API (including QuickConnect). This review focuses on correctness and safety across both modes, paying special attention to mobile-first constraints, multi-device concurrency, and the security/threat model. Key elements reviewed include the File Station transport assumptions, the lease/lock protocol with expected-old-ref checks, object-before-ref publishing order, caching behavior, and how the design adapts to mobile limitations (pure-JS Git engine without getBasePath(), ephemeral vs. persistent caches).

Overall, the design is ambitious and careful, explicitly aiming to prevent data loss, conflicts, or corruption. It addresses many typical pitfalls of sync and distributed version control. However, certain aspects remain high-risk if mis-implemented or if assumptions break (particularly in concurrent scenarios and on mobile). Some complex invariants may be challenging on resource-constrained devices, and the design’s correctness relies on careful coding of all safety measures. Below, we categorize issues by severity and discuss each in depth.


High-Risk Areas and Analysis

1. Multi-Device Concurrency & Data Consistency:
Risk: Concurrency issues in a multi-device environment can lead to data loss or split-brain scenarios if not handled correctly. Without safeguards, two devices syncing simultaneously could overwrite each other’s changes or produce divergent histories.

Design Mitigation: The Git-backed mode tackles concurrency with a two-tier safety approach: an advisory lock/lease file plus an expected-old-ref check on ref updates. When a client starts a sync, it attempts to create a lease file (.synology-sync/leases/.lock) on the NAS via File Station using an atomic “create if not exists” operation. If the lease is already present (another device is syncing), the second client must abort or retry later, preventing simultaneous writes to the same repo branch. This lock is not relied on as the sole guarantee: before pushing a new commit, the client re-reads the branch’s current HEAD OID and only updates the ref if it matches the “expected” old value it had when the lock was taken. This is essentially simulating an atomic compare-and-swap on the ref, akin to Git’s --force-with-lease safety. Objects are uploaded first, then the ref update is attempted last, under the lease, to ensure no partially-pushed commit becomes visible. If the ref has changed unexpectedly (i.e. some concurrent modification slipped through or occurred outside the plugin), the push is aborted; the client will detect a mismatch and must incorporate the new remote changes before trying again.

Assessment: This design correctly addresses concurrency: the lease file prevents concurrently running sync sessions from different clients, and the expected-old-ref discipline catches conflicts even if the lock didn’t (e.g. a malicious or non-cooperative client that bypasses the lock). If implemented exactly as designed, the risk of split-brain is low. The expected-old-ref check ensures one device’s changes are never overwritten by another’s without detection. The requirement to re-fetch/check the remote HEAD under the lock before ref update is crucial, and the design acknowledges it as essential (not optional) for consistency.

However, concurrency remains a high-risk area because it depends on multiple complex steps working properly. Potential failure modes:

Stale or Orphaned Locks: If a client crashes or loses connectivity while holding the lease, the lock file might not be deleted. Without an expiration mechanism, this could block all devices from syncing indefinitely. The design hints that the lease contains an expiry timestamp and an identifier (so presumably a timed lease) – this is good. It is critical that the implementation automatically handles stale locks, e.g. by allowing leases older than a threshold to be overridden or cleaned up, otherwise a single dead session could stall the system.
Simultaneous Lock Acquisition Edge Cases: The File Station API’s semantics for file creation need to be truly atomic. If two clients attempt to create the lock file at nearly the same time, one must fail cleanly. The design assumption (using force_parent=false to ensure no parent directory auto-creation) seems to rely on File Station returning an error if the file already exists – which should be atomic at a single-file level. But if File Station’s behavior is inconsistent or if a network glitch causes both clients to think they have a lock, there’s a slim chance of two concurrent writers. This is hard to fully eliminate, but the second-level expected-old-ref check would still prevent a “dirty write” (the second writer’s ref update will fail because the first writer’s ref already moved). The worst case in that scenario is the second device’s sync aborts (and must retry), not data corruption.
Coordination with Non-Plugin Actors: The design explicitly does not support other concurrent access methods (e.g. a user manually editing the NAS repo, or using a UNC share to simulate a working copy on multiple machines simultaneously). If a user bypasses the plugin’s controls (e.g. by mounting the NAS folder and editing it directly on two machines at once), all bets are off – conflicts or data loss can occur because those actions won’t respect locks or merge semantics1. The documentation should make it very clear that the Git mode requires each device to have its own vault clone, and users should not directly edit the bare repo or treat it as a normal folder.
In summary, concurrency is handled with robust mechanisms (lease + CAS update) that make the design sound, but the implementation must diligently maintain these invariants. Any bug in lock acquisition, release, or ref verification could be catastrophic (e.g. double-push or silent overwrite), so thorough testing (especially of interleavings and failure recovery) is imperative.

2. Overwriting Changes & Conflict Resolution (Both Modes):
Risk: When both local and remote have changed the same file between syncs, there’s risk of one side’s changes being lost or clobbered if conflicts aren’t detected and resolved properly. On mobile, with no persistent Git history, conflict resolution is particularly tricky (the plugin might not have a “merge base” commit to do a three-way merge in the traditional way).

Design Mitigation: In Simple File Sync mode, conflict handling was already implemented via a file-level strategy (likely using timestamps and generating conflict copies or tombstones when needed). The design commentary notes improvements like tombstone files and conflict resolution heuristics to prevent “ghost reappearances” or data loss, indicating a mature approach to one-side deletion vs modification issues2. However, file-level sync inherently has limitations: it’s essentially last-writer-wins unless special handling for conflicts is in place. The architecture calls this mode “Single User or simple workflows,” implying it’s not intended for heavy multi-device concurrent editing – which is wise, because it can’t do deep merge analysis and relies on simpler conflict policies.

In Git-backed mode, conflict handling is more powerful: since each change is captured as a commit, the plugin can detect conflicts via Git’s merge logic. The design explicitly forbids using a shared working folder (which would have conflated changes) and instead enforces separate clones per device so that merge conflicts can be detected via separate commit histories1. The plugin will attempt to auto-merge when possible: a clean text merge can be applied directly (with Git computing a combined result if changes don’t overlap)1. If a true conflict is detected (overlapping edits in the same lines), the design suggests not leaving raw Git conflict markers in the Markdown (which non-technical users might find confusing) but instead producing “conflict copies” or a guided merge note1. Essentially, when both sides made different changes to the same note, the plugin duplicates one version (or inserts human-readable conflict markings) rather than risk merging incorrectly. This approach sacrifices some elegance (users may have to manually reconcile two files) but preserves both versions of data – a safer default.

Assessment: The strategy strongly favors never silently overwriting divergent changes: either merge them automatically if non-conflicting, or surface them clearly via conflict notes. This is fundamentally sound for data safety. The risk remains that if conflict detection or staging fails, data could be overwritten by accident – but given the multiple checks (size, mtime, Git’s diff) and the bias toward preserving two copies, that risk appears low.

A potentially high-risk edge case to confirm is when a user edits a file during an ongoing sync. The release notes for file sync mode indicate this scenario was considered (mid-sync write protection and pre-download timestamp rechecks)2. For the Git mode, similar caution is needed: if a note file changes while the plugin is building a commit, does the plugin catch that and handle it (e.g. abort or delay commit)? The design implies the sync captures a “snapshot” of the vault before applying any remote changes (Step 4 of the flow ensures the local vault is snapshotted prior to checkout of remote changes) – so presumably it also ensures the commit uses a stable snapshot. If a file is modified by the user after that snapshot, ideally the commit logic would notice (for example by re-checking the file’s mtime or by validating the Obsidian vault “dirty” state) and abort or postpone committing that file until next cycle. If not, there’s a risk the commit might miss the latest edits. This is a corner case to verify in code.

3. Mobile Implementation Constraints (Ephemeral Git Repo & Cache):
Risk: On mobile (especially iOS), the plugin can’t use a native Git binary or a normal filesystem path (app.vault.getBasePath() is unavailable). The design’s solution is to use a pure JavaScript Git implementation (like isomorphic-git) working on an in-memory or sandboxed filesystem. This raises two safety concerns:
a. No persistent local Git history by default (initial implementation): If the in-memory Git repo is re-initialized on each sync run or app launch, the plugin might not have quick access to the last synced commit or the full object database between sessions. This could complicate merges and conflict resolution across app restarts.
b. Resource constraints: Doing Git operations on large vaults purely in JS (in a mobile WebView) may be slow or memory-intensive. For correctness, if the process takes too long or uses too much memory, some mobile OSs might kill the background task or the app — potentially mid-sync, which could leave a lock file stranded or a partially uploaded repository.

Design Mitigation: The roadmap notes a future plan to implement a persistent mobile Git cache (e.g. using @isomorphic-git/lightning-fs and IndexedDB) so that mobile devices can reuse a local repo state across sessions, avoiding re-downloading everything every time and enabling proper three-way merges. But currently this is to be implemented later. For now, on iOS the plugin uses an in-memory approach per sync. The design partially compensates by making each sync cycle self-contained and safe to restart: since it always fetches the remote’s latest state after acquiring a lock and then compares it to a fresh snapshot of local files, it won’t miss remote changes just because it lacks persistent history – it just may have to download more data to determine differences.

Assessment: Correctness is preserved on mobile, but this approach trades efficiency for safety. Not having a persistent local repo means the plugin can’t rely on simply diffing against a local copy of the last HEAD – instead it must fetch enough remote data each time to compute differences. This is heavy but safe; the client will always detect differences by brute force if needed (e.g. computing file hashes or re-downloading relevant objects if modification is suspected). The design calls for a minimal object fetch (not a full clone) – e.g., just the commit and tree objects needed to compare state1 – but if packing or caching complicates that, the plugin might conservatively fetch more. The safety principle is to “fail closed” if anything is uncertain: e.g., if a required object is in a packfile and not easily accessible, it should fetch the pack or drop down to a full sync rather than risk missing data. This is a sound approach: it ensures the worst-case scenario is extra network usage or a slower sync, not a corrupted or out-of-sync repository.

However, there is a medium-risk consequence: with no persistent merge base, every potential conflict on mobile might get resolved via conflict copies rather than an elegant automatic merge. Without storing the common ancestor commit, the plugin may not truly perform a three-way merge of diverged changes – it might simply see “file X changed locally and remotely; we have no easy base snapshot, so mark conflict.” This conservative strategy avoids data loss but could lead to more frequent conflict files for mobile users. Recommendation: If not already planned, consider storing the last-synced commit ID (and perhaps the base state of each file or a vault manifest hash) persistently on the mobile device. This would allow the plugin to identify a merge base and attempt a true 3-way merge of text changes, reducing unnecessary conflicts while still being safe.

Additionally, the plugin should guard against partial syncs on mobile. If the app is backgrounded or closed mid-sync (common on mobile), ensure the plugin either (a) quickly acquires the lock and finishes critical sections, or (b) robustly cleans up or marks an incomplete sync so that next run starts fresh. The lease mechanism’s use of an expiry time is key here – a terminated mobile sync must not leave behind a permanent deadlock. It might be wise to use relatively short lease expirations on mobile, given the higher chance of abrupt termination.

4. File Station / QuickConnect Transport Assumptions:
Risk: File Station is not a native Git transport, so it cannot use Git’s built-in network protocols or locking. It’s essentially a REST API for file manipulation. This means:

Operations might not be as atomic or isolated as with a proper Git server.
Metadata like timestamps and file sizes might not correlate strongly with content (especially if File Station repacks files or compresses them, or if network latency triggers timeouts).
The plugin must handle network failures or partial transfers carefully, because the File Station API might not have built-in transactionality.
Design Mitigation: A lot of the complexity in the Git-backed mode is to simulate the atomicity of a Git push using File Station. The plan to always upload new objects before pointing the branch ref to the new commit ensures that if a sync fails mid-way (e.g., connection drops after some objects uploaded but before the ref is set), the remote repository remains on the old commit. Those uploaded objects will be unused (dangling) and can be pruned later, but they won’t cause inconsistency. Meanwhile, the lease ensures that only one set of new objects is being uploaded at a time, minimizing the possibility of remote repository corruption. After any failure, the lease holder will either gracefully release the lock or the lock will expire, allowing a retry.

Additionally, to cope with File Station’s limitations, the design says that after obtaining the lock, the client re-fetches/validates any needed remote data. It avoids trusting stale caches: for example, the client won’t just rely on a previous local idea of HEAD if that wasn’t freshly checked under the lock. This ensures that after acquiring the lock, the client’s knowledge of remote HEAD is authoritative.

Assessment: The File Station/QuickConnect approach is inherently more complex than using a Git server or direct FS, but the design seems aware of all necessary precautions: atomic sequence, explicit verification, and fallback strategies. One medium-risk assumption to double-check is the reliability of File Station’s file operations for creating and updating the ref file (likely the file representing refs/heads/main) with strict ordering. It must not be buffered or cached in a way that a partially written ref could appear – presumably the plugin writes the ref file with an API call that either succeeds or fails wholly. Given these constraints, the design’s approach is solid. Nevertheless, extensive testing of failure scenarios (network drop, File Station timeouts) is warranted to ensure the plugin always fails safe.

Another risk (though more performance than safety) is if QuickConnect misbehaves or chooses a suboptimal path. The architecture describes a state machine for QuickConnect (with phases like resolve_candidates, retry_without_token, OTP handling, etc.). If those flows have bugs, a sync might fail to connect or inadvertently drop an existing session. From a safety perspective, that’s not data-corrupting – it just results in inability to sync. It’s more of a reliability problem. So it’s lower priority for safety but still important for user experience.


Medium-Risk Areas and Analysis

5. Complexity of Cache and Incremental Transfer Logic:
Risk: The Git-backed mode aims to be efficient by doing smart, incremental syncs (fetching only needed objects rather than full re-clones). This logic – identifying exactly which tree and blob objects to fetch based on differences – is complex. If there is any flaw, the plugin might either download too little (missing data needed for comparison) or trust a cache that is stale. A worst-case bug here could cause missing updates (e.g., not pulling a remote change due to mistakenly thinking nothing changed) or divergent repository state.

Design Mitigation: The design repeatedly emphasizes caution over cleverness: it says any metadata shortcut (like using file timestamps or cached object IDs) must be “conservative” – if there’s any doubt, it will verify or transfer rather than assume. For example, it might use File Station’s file size and mod-time to guess if an object changed, but if it can’t prove it’s unchanged, it will re-download it1. Similarly, packfiles are treated delicately: if the needed objects are stored in a pack on the server, the plugin either needs to parse those packs (potentially complex) or fallback to a broader fetch. The design explicitly says to prefer failing closed (not proceeding with incomplete data) over guessing and possibly missing something.

Assessment: The risk here is largely mitigated by the safety-first approach. The consequence of any uncertain scenario is extra work (download the whole pack or the entire repo) which does not compromise correctness, only performance. So while the incremental sync logic is intricate and prone to edge cases, the worst likely outcome of a bug is a slow sync or an unnecessary conflict copy, not silent data divergence, provided the “safe fallback” paths are properly implemented. A medium risk remains that the complexity might hide a subtle bug or corner case – but these would likely manifest as obvious sync failures (which users can report) rather than undetected data loss. Logging should be thorough in this area to help identify if the plugin ever decides to “skip” something when it shouldn’t.

6. Mobile’s Ephemeral Git Repo & Conflict Handling (continued):
We discussed above that mobile’s lack of persistent git storage is a design choice with some downsides. To expand on correctness: if the plugin doesn’t store the last known local commit (or base snapshot) persistently, it might have to treat any scenario with concurrent updates as a full conflict (instead of using Git merge). This means users may see more conflict copies on mobile in multi-device scenarios. While not a data loss risk (both versions are kept), it can be confusing or annoying to users. It’s a medium risk in terms of user trust in the system’s correctness (they might think “why do I keep getting conflicts?”). The design’s planned improvement (persistent on-device storage using IndexedDB) should alleviate this by enabling true merges and faster diffs. Until then, it’s important that documentation sets the expectation: multi-device simultaneous editing is better handled by the Git mode on desktop or with careful sequencing on mobile, due to this limitation.

7. Simple File Sync Mode in Multi-Device Use:
While the Git mode is clearly designed for multi-device concurrency, some users might still attempt to use Simple File Sync across two devices (because it’s simpler to set up). This mode has tombstones and conflict strategies, but it is fundamentally less robust than Git for concurrent edits. The architecture acknowledges it’s for “single user or simple workflows.” If used beyond that scope, medium-risk problems like ghost notes (one device re-creating a note deleted on another), or undetected conflicts (last writer wins scenarios) can occur. The latest improvements like cross-device tombstone purging2 mitigate some of these, but it’s still advisable to warn users that if they need real-time multi-device sync, the Git mode is safer. This is less a design flaw and more a documentation/UX concern: the plugin should funnel multi-device users toward Git mode, or at least explain the limitations when multiple devices use Simple Sync.

8. Handling Large Files or Repositories:
If the vault contains very large files (attachments) or a very deep history, there’s a risk of hitting timeouts or memory limits, particularly on mobile. This might cause incomplete sync runs. However, the design’s incremental approach should limit how much data is transferred per sync (only changes, ideally). On first sync initialization with a lot of data, risk of failure is higher (e.g., getting everything into a commit might be heavy for in-memory on mobile). This is a corner case but worth mentioning as a medium risk to test, because if an initial mobile sync fails repeatedly due to resource limits, it could strand a user with a half-initialized vault. Possibly the plugin should encourage doing the first full sync on a more robust device or ensure a chunked approach for initial seeding of a big vault on mobile.


Low-Risk or Minor Issues

9. Security of Data in Transit & at Rest:
The plugin uses Synology’s official API (File Station) over either local network or QuickConnect (which presumably tunnels via HTTPS through Synology’s relay). This is fine for confidentiality and integrity as long as the NAS credentials and QuickConnect token are handled properly. The design outlines a state-driven login flow with OTP handling and token persistence. There’s no mention of encryption beyond what File Station/QuickConnect provide (which is acceptable – they rely on HTTPS). A threat here is if the credential (session token or password) is compromised – an attacker could impersonate the sync client. The plugin can’t fully mitigate that beyond storing tokens securely (which hopefully it does e.g. using Obsidian’s settings store) and perhaps offering OTP/2-factor for logins (which they do support in the flow). This is a low risk inherent to any cloud connectivity. As a precaution, logs should avoid printing sensitive info like passwords or tokens. The design’s debug states (like persistedTokenAction) should be careful not to leak actual auth tokens in logs.

10. Malicious or Buggy Client Behavior:
If a malicious or misbehaving client (perhaps a modified plugin or an external script with credentials) tries to subvert the protocol – e.g., by writing to the bare repo without taking a lease or by arbitrarily rewriting history – the design’s defensive checks help limit damage. For example, an external process could theoretically push a new commit directly via UNC path (if they have access to the NAS) without a lock. If that happens concurrently with a plugin sync, the plugin’s expected-old-ref check will stop it from blindly overwriting (the ref on NAS would have changed unpredictably, causing a mismatch). The worst-case scenario from such an adversary is they succeed in pushing a commit out-of-band, which means the next plugin sync will treat it as just another remote update to merge. This isn’t desirable but is not catastrophic. If an external client holds the lease file maliciously (creating it and never releasing), that’s effectively a denial-of-service to the sync. The inclusion of a timestamp and expiry in the lease file can mitigate that: the plugin can eventually decide “this lock is expired, I will override it.” There should be code to break an expired lease (possibly with a warning in logs or UI). Provided that’s implemented, a malicious stale lock can be cleared after some timeout.

11. Admin/Developer Direct Repository Access:
The architecture explicitly notes that direct access to the bare repository via a mount or UNC path is not part of the supported user sync path, and if done, it must honor the same safety rules (lease + expected-old ref) to avoid interfering with the plugin1. This scenario is more about someone with advanced knowledge trying to script a backup or run git gc or similar on the NAS repo. If they do so without coordination, they could cause race conditions or even repository corruption (e.g., running git gc could repack objects while the plugin is reading them). This is considered a low risk in practice because it’s an advanced scenario; still, we might recommend documenting how to safely do maintenance on the repo if needed (e.g., take plugin offline, ensure no active leases, then proceed, etc.).

12. Privacy and Logging:
The architecture mentions including runtime/platform info in logs (like user agent, device type, a hash of the host name) for support/debugging, while avoiding logging actual hostnames or file paths (for privacy). This is a good practice. One minor risk is to ensure no sensitive data from notes or credentials gets inadvertently logged, especially if users might share logs for support. The design already seems mindful of this (with host “fingerprints” instead of actual names).


Threat Model Considerations

This section addresses specific threats and how the design handles them:

Malicious or buggy plugins/clients: A rogue client that doesn’t follow the protocol (e.g. ignoring locks or expected-old-ref) could attempt concurrent writes or rewriting history. The combination of lease enforcement (others can’t get the lock) and ref verification (others can’t push without up-to-date base) provides good protection against such a client polluting or overtaking the sync. In effect, if a malicious client tries to “race” a well-behaved one, it will likely either fail to acquire the lock or fail the ref update (unless it times its push exactly between the windows, which is very narrow). A malicious client could also spam the bare repository with extraneous objects (since File Station just deals with files) – e.g., uploading bogus packfiles. This wouldn’t affect the sync logic (the plugin only cares about the HEAD commit and reachable objects), but it could bloat the repo storage. Mitigation is mostly by limiting who has credentials; detecting such conditions automatically is hard, but if the plugin notices a huge number of unreachable objects or an abnormal repository state, it could warn the user.
Compromised NAS credentials: If an attacker obtains the user’s NAS login or the plugin’s stored token, they could potentially read or modify the vault contents via the API. This threat is largely outside the plugin’s direct control – standard advice applies: use strong passwords, enable 2-factor (which the plugin does handle via OTP), and possibly scope a separate NAS account for the plugin with limited access. The plugin’s threat model should assume the NAS endpoint is secured by authentication and focus on preventing unintentional issues rather than targeted attacks. However, one relevant scenario is if an attacker with credentials purposely tries to interfere (like creating a fake lock file or pushing a conflicting commit). As discussed, the design’s controls would limit damage to denial-of-service (lock jam) or at worst causing a conflict that the real user must merge.
Stale or replayed leases: A stale lease (due to crash or malicious holdover) could prevent progress. To counter this, leases include an expiry. The plugin should treat any lease older than a safe threshold as invalid and allow a new sync to proceed (possibly after logging a warning). A “replayed” lease (someone copying an old lock file into place) would be seen as expired because the timestamp would be old. There’s a slight chance of clock skew issues (if client and NAS clocks differ widely), but since the plugin itself sets and checks the lease timestamp, it can treat time relative to start-of-sync to measure expiration. Also, the lease file might include a unique token or client ID – not explicitly confirmed, but if so, a new client can detect “this lease isn’t mine, belongs to X, from time Y”.
Partial uploads and aborted syncs: The object-before-ref approach ensures partial commits don’t break consistency: if a client uploads some new objects but never updates the ref, other clients will ignore those objects (they aren’t reachable from any branch). The biggest risk is wasted space or a temporarily slow subsequent sync (the next sync might see the needed objects are already there and skip uploading them again, or it might re-upload if it doesn’t trust them). The design explicitly calls for retrying safely after partial publishes. As a further mitigation, perhaps the plugin could detect an incomplete prior attempt by noticing the remote HEAD is still old but some objects of the supposed new commit exist – if that scenario is detectible, it could avoid re-uploading identical objects. However, even if it doesn’t, re-uploading is harmless aside from performance. So partial upload is not a correctness threat, just something to handle gracefully for user experience (ensuring the next try will succeed or at least not hang).
Packfile edge cases: Synology’s bare repo may accumulate packfiles due to normal Git operation (especially if any maintenance or a non-plugin push occurred). The plugin may need to read packfile indexes to fetch specific objects. A malicious scenario is if a packfile is somehow corrupted or has unexpected format – the plugin must not assume pack contents blindly. The plan to “fail closed” (if required objects are not found as loose files, either parse the pack carefully or fetch a larger set) is prudent. If the plugin can parse pack indexes via pure JS (isomorphic-git might have support), that could be done; if not, a conservative fallback is to fetch the entire pack or even run a full “clone” of the bare repo to local memory. That’s slow but safe. The main threat of packfiles (corrupted or manipulated pack contents) is similar to any Git scenario – generally, if you trust your storage (NAS), the risk is low. If an attacker could replace a pack with one containing malicious data, they could cause the plugin’s Git library to crash or conceivably run into pathological behavior. This is fairly theoretical; normal repository corruption due to a bug or storage fault is more likely. We might consider verifying the checksum of packfiles, but since that’s what Git does anyway, using Git objects ensures each object is cryptographically hashed (SHA-1 or SHA-256 depending on Git version).
Admin direct-bare access: Covered above; the threat is mostly inadvertent – an admin doing maintenance or a user doing something outside the plugin could break assumptions. The design recommends any such access must follow the plugin’s rules (obtaining the lock and updating with expected-old semantics)1. In practice, a user messing with the bare repo is out of scope for plugin code (we can’t prevent it), so at best the documentation must strongly discourage it unless the user is an expert. Perhaps the plugin could put a README in the bare repo that warns “This is an Obsidian Sync repo – do not edit directly.”
Privacy and logging: The design’s logging approach has been addressed as low risk. If there is any threat, it’s if logs inadvertently leak note content or identifying info. The design already says to avoid raw pathnames; using a hash fingerprint is a good approach. This threat is minimal if the guidelines are followed.

Recommendations & Alternatives

The current design is thorough and addresses core safety concerns. However, a few recommendations arise where the design might be improved or simplified:

Explicit Handling for Stale Locks: Implement a clear strategy for stale lock files. For example, set the lease expiration to a reasonable window (maybe a few minutes) and have the plugin automatically break and replace a lease older than that. Provide a log or UI message when this happens (so a user knows if a prior session crashed). This ensures a crash or a network partition doesn’t halt sync indefinitely.
Mobile Persistent Cache (Future Priority): As the design already notes, implementing a persistent cache (via IndexedDB or similar) on mobile should be a priority for a future version. It will not only improve performance but also reduce conflict frequency by enabling true merge logic. If persistent storage is limited or complex, even storing just the last synced commit ID and a simple mapping of file paths to last-synced hash could help as an interim measure. That way, the plugin could attempt a three-way merge by reconstructing needed base version from the remote if not locally cached.
Conflict Handling Experience: The design’s conflict approach is safe, but consider making the conflict presentation as user-friendly as possible. Instead of simply duplicating files with “conflict copy” names, perhaps present a single “Conflict in [Note Title]” note that summarizes the differences or points the user to the two versions. This might be more of a UX feature request, but it can prevent confusion when conflicts do occur, especially for non-technical users who might not immediately understand why there are two copies of their note.
Verify the File Station API details: The team should double-check how File Station’s API behaves for file creation (for locks) and file modification (for the ref update). If File Station doesn’t support conditional operations (like an HTTP If-Match equivalent), consider implementing a simple alternative: e.g., writing the new ref file to a temporary name then renaming it if the old ref file content matches expected (some caution: the API might not allow such conditional logic, hence the explicit expected-old in content might suffice). If the API has any eventual consistency quirks, maybe insert small delays or re-check reads after writes. The design is correct conceptually but depends on API subtlety, so it’s worth prototyping these sequences under stress.
Graceful Degradation of Incremental Sync: If the logic for minimal object sync proves too brittle or complex, a safer (but slower) fallback could be to allow a configurable “full sync” mode – essentially always re-pull the entire repo state, or large parts of it, when issues are detected. The design already says to fetch a broader object set if needed, which is good. Just ensure that code path is well-tested (simulate a scenario where every sync acts like a cold start). This could also be triggered by a debug switch so that if a user reports weird sync issues, they can be advised to turn on “safe sync mode” which, for instance, disables reliance on caches.
Documentation & User Guidance on Mode Selection: Emphasize in the UI and docs when to use each mode. Possibly rename “Simple File Sync (Single User)” to “Single-Device File Sync” to make it even more apparent that it’s not meant for multiple simultaneous devices. Conversely highlight the Git-backed mode as “Multi-Device (Git-based) Sync” or similar, and mention that it’s intended for multiple devices or advanced use. This reduces the chance that users pick the wrong mode for their scenario.
Terminology and Complexity in Documentation: The architecture description is very detailed – possibly too detailed for the average user. For the README (user-facing), consider simplifying: you can describe Simple Sync as “stores your notes as regular files on the NAS (so you can view them via File Station, but limited conflict handling)” vs Git Sync as “uses an internal version control system (Git) to manage changes, which is more robust for multiple devices but stores your notes in a special format on the NAS (not directly human-readable on the NAS, but fully accessible through Obsidian on each device).” Non-developer users may not know what “bare repo” means, so it helps to phrase it as “a special kind of folder on your NAS that acts as a central hub for syncing, but won’t look like your notes if you browse it directly.” Save the detailed explanation of Git internals, commit graphs, lease files, etc., for developer documentation or an extended manual.
Clarify Invariants in Contributor Docs: For future maintainers, clearly mark which steps are critical for safety. The architecture write-up is excellent, but it’s long. It might help to summarize must-follow rules (like a checklist of invariants): e.g., “Always acquire lock before reading or writing remote. Always verify ref consistency before writing. Always upload all objects before moving the ref. Always snapshot local vault state before writing any remote changes to it,” etc. Some of these rules come up in the text; highlighting them explicitly in a quick-reference section could prevent mistakes in future code changes.
Simpler Alternative (Consideration): One alternative design approach (though probably rejected for performance reasons) would be to leverage the Synology Git Server or a separate lightweight Git server accessible via SSH/HTTP. This would offload all concurrency and pack-handling complexities to Git’s built-in server logic. However, as the design notes, requiring Git Server on the NAS undermines the “no extra Synology packages” requirement, and QuickConnect doesn’t natively support Git protocol. So this alternative is not aligned with the product goals and was rightly set aside.
Another Alternative – conflict-free replicated data types (CRDTs): Well beyond the current scope, but for brainstorming: some modern real-time sync solutions use CRDTs to merge text without version control. That would be a radical departure and unnecessary here given the robust Git approach. Git is a well-understood strategy that brings auditability and history, which is a plus for note versioning.
In summary, the recommendations above aim to fortify the existing design rather than replace it, as the architecture is fundamentally solid for correctness and safety.


User & Contributor Documentation Guidance

For a plugin of this complexity, clear documentation is critical. Below are some suggestions to improve user-facing README.md and contributor docs (e.g. CONTRIBUTING.md) to prevent misunderstandings and unsafe usage:

Explain the Two Sync Modes in Simple Terms: The README should start with a non-technical explanation of Simple File Sync vs Git-backed Sync. Focus on what the user cares about:
Simple File Sync (Single Device): Good for one device or very simple use; your notes are stored as regular files on the NAS so you can browse them directly via the NAS interface. If two devices use this mode at the same time, they may encounter conflicts or duplicated notes.
Git-backed Sync (Multiple Devices): Intended for multiple devices or collaborators; uses a version control repository on the NAS to merge changes. This is more resilient to simultaneous edits and keeps a history of changes. However, the notes on the NAS are not directly human-readable (they’re stored in a Git repository format).
Use language that avoids jargon like “bare repository” or “ref updates” when addressing end users. Instead of “bare Git repo,” say “a special storage format (a Git repository) that the plugin manages for you.” Emphasize that users do not need to know Git to use it – the plugin handles everything – but they should be aware of the trade-offs (especially that they can’t directly browse the synced folder on the NAS in Git mode).
Highlight Mode Selection Guidance: Provide a clear decision matrix or a few example scenarios in the README for which mode to pick. For example: “If you only use Obsidian on one computer and want easy access to your notes via your NAS’s file browser, use Simple File Sync. If you use Obsidian on multiple devices or worry about conflicting changes, use Git-backed Sync for better conflict resolution.” This will steer users appropriately and reduce misuse.
Warn Against Unsupported Use Cases: Clearly state that opening the NAS’s Git repository folder in Obsidian or editing it manually is not supported and can cause conflicts or data loss. Encourage users to always use the plugin to sync, not external manual file operations, especially in Git mode. That means if they want to access their notes on another device, they should set up the plugin there (with the same mode and repository) rather than, for example, using the Synology Drive client concurrently on the same folder.
Document the Safety Guarantees & Limitations: In user-friendly terms, mention the protective measures:
Only one device can sync at a time (to protect data). If two devices try to sync at once, one will wait or retry.
The plugin keeps an offline history of changes (for Git mode) so you can recover old versions if needed (this is a selling point for power users).
If two devices edit the same note at the same time, the plugin will not silently merge them (to avoid errors); instead, it will save both versions so you can manually merge them (and possibly in the future might support more automatic merging).
Point out that because of these measures, data loss is very unlikely – the worst-case scenario is you’ll have to resolve a conflict manually, but your data will be safe. This can reassure users who fear using a third-party sync solution.
Contributor Documentation (CONTRIBUTING.md): The architecture discussion belongs in a more technical doc for contributors. Ensure that file includes the details about how the sync works under the hood, the invariants (like lock usage, expected-old-ref), and instructions for running tests. Perhaps include a succinct list of “design invariants” as mentioned, and note which aspects are critical to get right (e.g. “Never update a branch ref on NAS without verifying the expected old value – see GitService.push() for how this is done”). Also, document any tricky platform-specific code paths: e.g., how the mobile in-memory git differs from desktop (so future contributors don’t inadvertently break mobile).
Clarify Terminology: Some terms used internally might confuse new contributors or advanced users. For instance, "lease" might be better understood as "lock file", and "expected-old-ref" can be described as “the commit ID that the remote repository is currently at, which we expect to still be the same when updating – to avoid conflicts”. Ensure terms like “bare repo”, “worktree”, “packfile” are explained or a reference is given, at least in contributor docs, since not everyone is a Git expert.
Safety Warnings for Advanced Usage: If the plugin allows configuration changes (like including .obsidian folder in sync, or fiddling with conflict strategies), point out the implications. For example, synchronizing the entire .obsidian folder might cause Obsidian settings to clash across devices (since some settings are device-specific). Make sure the README or settings UI labels highlight what default is safest (e.g., “By default, only your notes are synced – not your workspace settings or community plugins, to avoid conflicts. You can opt in to sync these if you know the risks.”).
Error and Recovery Guidance: Provide guidance in docs or a FAQ for what to do if something goes wrong. E.g., “If your sync gets stuck (maybe due to a network issue or a stuck lock), here’s how to recover: [steps to remove a stale lock file, or re-initialize a sync].” This will help users handle rare but possible situations like an unbreakable lock (maybe if the lease expiration mechanism fails or if clocks are misaligned).
Tone and Clarity: The user-facing docs should be simplified. The architecture description given is highly technical (which is great for an internal design doc), but the README should be pruned to the essentials from a user perspective: what the plugin does, how to set it up (with step-by-step, including creating a bare repository on the NAS if needed), what each mode means, and common gotchas (like not mixing Sync with direct editing of the same folder outside Obsidian). More intricate details (like internal flow charts or justification of design decisions) are better placed in an “Architecture” section of the contributor guide or a wiki for those interested.
Essential vs. Advisory Invariants: Finally, be explicit (especially in contributor docs) about which protocols are non-negotiable for correctness and which are merely best practices or future enhancements. For example, “Only update the remote ref with a known expected-old value” and “Only release the lease after finishing all remote writes (including ref update)” are essential invariants for correctness. In contrast, something like “Encourage direct NAS Git users to follow the lease protocol” is advisory – it’s good to document and recommend, but the plugin cannot enforce it. Calling out the difference will help maintain focus on what absolutely must remain true for data safety.


Conclusion: The proposed architecture is quite comprehensive in addressing safety and concurrency, especially for the Git-backed mode which appears well-suited for multi-device sync with minimal data loss risk. The Simple File Sync is acceptable for single-device use but should be clearly documented as such. The highest risk lies in implementing all these safeguards correctly, especially under the constraints of mobile environments and the quirks of the File Station API. With careful testing (including simulated crashes, network failures, and concurrent scenarios) and by following through on planned improvements (like persistent mobile storage), the design should achieve its goal of robust, safe synchronization of Obsidian vaults to a Synology NAS.

ARCHITECTURE.md
```diff
@@ -20,11 +20,11 @@ Behavior:
 - The NAS folder contains normal Markdown/assets that can be browsed directly.
 - Conflict handling is the plugin's file-level conflict strategy.
 - This is the default and mobile-friendly mode.
-- This mode is intended for single-user or simple workflows.
+- This mode is intended for single-user or simple workflows. The settings UI and README should steer multi-device or concurrent-editing users toward Git-bare-backed Sync instead.
 
-### Git-backed sync over File Station / QuickConnect
+### Git-bare-backed Sync
```
VertigoRay: `### Git-bare-backed Sync (Multi-User)`

Initial references, like on L3 should also, specify `Multi-User`.

ARCHITECTURE.md
```diff
@@ -1,12 +1,12 @@
 # Architecture
 
-This plugin supports File Station connections via QuickConnect or a direct Synology address, with two intentionally different sync architectures layered on top: simple file sync or Git-backed sync over File Station. Settings, validation, logs, and support guidance should preserve this distinction instead of treating all modes as variants of one remote folder sync.
+This plugin supports File Station connections via QuickConnect or a direct Synology address, with two intentionally different sync architectures layered on top: Simple File Sync (Single User) or Git-bare-backed Sync. Settings, validation, logs, and support guidance should preserve this distinction instead of treating all modes as variants of one remote folder sync. Code should enforce the same boundary with separate sync engines for Simple File Sync (Single User) and Git-bare-backed Sync; shared modules should be limited to File Station transport, diagnostics, and other mode-neutral utilities.
```
VertigoRay: `Git-bare-backed Sync (Multi-User)`

Initial references, like on L3 should, specify `Multi-User`.

ARCHITECTURE.md
```diff
@@ -39,10 +39,14 @@ Behavior:
...
+- Native Git is an implementation detail against the local worktree/cache/bare mirror. It is not a supported remote transport surface.
+- UNC and `/mnt` direct-bare access is admin/developer tooling outside the mobile plugin contract. If such tooling writes to the bare repo, it must honor the same lease and expected-old-ref discipline or it can race plugin clients.
+- Git SSH/HTTPS remotes are intentionally not supported by this plugin. Users who want normal Git remotes should use a pure Git/Obsidian Git plugin instead.
 - The picker/validation must only accept a bare Git repo shape: `HEAD`, `objects/`, and `refs/`.
-- On desktop runtimes with local filesystem access, this mode may use native Git against the local checkout.
-- On iOS/mobile runtimes without `getBasePath()`, this mode uses a pure-JS Git engine over Obsidian's vault APIs instead of desktop-only Node/native Git APIs.
+- On desktop runtimes with local filesystem access, this mode may use native Git against the local checkout/cache while still treating File Station/QuickConnect as the supported NAS transport.
```
VertigoRay: Let's be more clear:
`On device runtimes (Obsidian plugin) with local filesystem access as an (such as desktops, laptops, etc.), this mode...`

_Feel free to improve grammar for readability._

ARCHITECTURE.md
```diff
@@ -39,10 +39,14 @@ Behavior:
...
+- On iOS/mobile runtimes without `getBasePath()`, this mode uses a pure-JS Git engine over Obsidian's vault APIs instead of desktop-only Node/native Git APIs. Mobile Git support must document practical repository/cache limits and fail with actionable guidance when repo size, packfile size, or storage quota exceeds what the runtime can safely handle.
```
VertigoRay: Same as comment on L48. I like that we reference `getBasePath()` here, maybe we should on L48 as well, if that's the deciding factor...

ARCHITECTURE.md
```diff
@@ -57,30 +61,101 @@ Bootstrap requirements:
...
-- Preserved conflict-copy names must be stable for the same local content so repeat syncs do not create unbounded duplicate copies. New local content should still get a distinct preserved copy.
+- Preserved conflict-copy names must be stable for the same local content so repeat syncs do not create unbounded duplicate copies. New local content should still get a distinct preserved copy. Conflict presentation should be user-friendly and preserve both versions; raw Git conflict markers should not be the only recovery surface for non-technical users.
+
+#### Git-bare-backed Sync flow
+
+The Git-bare-backed Sync flow has four distinct jobs: acquire the NAS sync lease, read/fetch remote Git refs and required objects from Synology into the Git cache, reconcile the local vault snapshot against that state, and publish any resulting Git commit back to the NAS bare repository before releasing the lease. Reading remote Git state must not materialize or check out remote files into the readable vault before the local vault snapshot. A successful mobile/local-write sync must include all four when local notes changed.
```
VertigoRay: For NAS operations, let's avoid git terms like fetch/pull/push/clone. For NAS operations, use terms like read/sync/write. This removes ambiguity. Let's define the terms at top of sync section, before both simple and git-bare-backed sections.

ARCHITECTURE.md
```diff
@@ -57,30 +61,101 @@ Bootstrap requirements:
...
+    O --> R
+    R --> Z
+    Z --> S
+```
+
+Operational sequence:
+
+- **Step 1: Create/acquire the NAS sync lock first.** Before doing sync work, create/place the File Station lock/lease on the NAS. If another client already holds it, abort/retry rather than racing. The lock remains held until publish succeeds or the run safely aborts.
+- **Step 2: Local cache/mirror HEAD preflight.** If a local Git cache/mirror exists, read its HEAD/ref and object availability as an optimization/readiness check. This can identify an obvious no-op or missing-object condition, but it is not final authority for writes.
+- **Step 3: Read NAS HEAD/ref through File Station.** The product authority is the NAS bare repo as observed through File Station/QuickConnect, not SSH/HTTPS Git remotes and not a stale local cache. This step may fetch required refs/objects into the Git cache, but it must not check out or materialize remote files into the readable Obsidian vault.
```
VertigoRay: Talking about SSH/HTTPS feels redundant here. We've already noted this got protocols are not supported in this plugin.

ARCHITECTURE.md
```diff
@@ -57,30 +61,101 @@ Bootstrap requirements:
...
+This plugin is not a general-purpose Git client and should not grow a separate first-class path for non-File-Station Git remotes. Existing Obsidian Git plugins already serve normal Git remotes well. The Git-bare-backed Sync mode here exists specifically to use Synology File Station / QuickConnect as the transport.
 
 ## Settings UI contract
 
 The settings UI should make the choice explicit with a Sync mode selector:
 
-- Simple file sync (single user) shows Remote folder path.
-- Git-backed sync over File Station / QuickConnect shows NAS bare Git repo path.
-The settings UI should not present non-File-Station Git remotes as a primary sync mode.
+- Simple File Sync (Single User) shows Remote folder path.
```
VertigoRay: In the UI, show `Single User (Simple File Sync)`

ARCHITECTURE.md
```diff
@@ -57,30 +61,101 @@ Bootstrap requirements:
...
+- Git-bare-backed Sync shows NAS bare Git repo path.
```
VertigoRay: In the UI, show `Multi User/Device (Git-bare-backed Sync)`

ARCHITECTURE.md
```diff
@@ -57,30 +61,101 @@ Bootstrap requirements:
...
+The settings UI should not present non-File-Station Git remotes as a primary sync mode. It should also explain the mode choice in user terms: Simple File Sync (Single User) stores readable files on the NAS but is limited for concurrent multi-device use; Git-bare-backed Sync is intended for multi-device safety but stores a Git repository on the NAS rather than directly browsable notes.
```
VertigoRay: Great description, but note my L154 and L155 comments.

VertigoRay (commented):
Make corrections and merge it.

Issue #118, opened 2026-05-28
Research review: implementation strategy, correctness envelope, and test gaps for obsidian-synology-sync


VertigoRay:
This is a research-driven review of the current implementation strategy in `obsidian-synology-sync`. The repo is moving quickly — the GitHub metadata currently shows an active `src/` and `tests/` layout, 68 commits, and 35 releases, with a recent release only days old — which is a good sign for iteration speed, but it also raises the bar for architectural discipline, regression coverage, and doc consistency. The plugin is explicitly still marked **BETA** and claims support for both desktop and mobile Obsidian clients. [1](https://github.com/VertigoRay/obsidian-synology-sync)[2](https://github.com/VertigoRay/obsidian-synology-sync/blob/main/README.md)

My overall conclusion: the repo shows **strong product thinking and a notably safety-conscious architecture**, especially around the split between simple file sync and Git-bare-backed sync, but the implementation strategy has crossed from “file sync utility” into “distributed systems product.” That means correctness, integration testing, failure injection, and support boundaries need to be upgraded to match the complexity that now exists in the design. [3](https://github.com/VertigoRay/obsidian-synology-sync/blob/main/ARCHITECTURE.md)[4](https://github.com/VertigoRay/obsidian-synology-sync/issues/31)

## Review method / evidence base

This review is based on the repository README, architecture doc, changelog, issue history, release notes, and indexed GitHub metadata, plus upstream Synology, Obsidian, and Git documentation where relevant. In other words: this is not just a feature review; it is an implementation-strategy review grounded in the repo’s own stated invariants and change history. [1](https://github.com/VertigoRay/obsidian-synology-sync)[2](https://github.com/VertigoRay/obsidian-synology-sync/blob/main/README.md)[3](https://github.com/VertigoRay/obsidian-synology-sync/blob/main/ARCHITECTURE.md)[5](https://github.com/VertigoRay/obsidian-synology-sync/blob/main/CHANGELOG.md)[6](https://github.com/VertigoRay/obsidian-synology-sync/releases)[7](https://global.download.synology.com/download/Document/Software/DeveloperGuide/Package/FileStation/All/enu/Synology_File_Station_API_Guide.pdf)[8](https://docs.obsidian.md/Plugins/Vault)[9](https://git-scm.com/docs/git-update-ref)

### 1) The architecture split is the right call

[...]

PR #120, opened 2026-05-28
Harden Git File Station lease and ref publication

VertigoRay:
## Summary
- recover only expired Git-over-File-Station leases after reading lease metadata
- verify acquired lease metadata/token before publishing
- publish Git objects before branch refs, then publish the branch ref via temp upload + File Station rename when available
- re-read/verify the branch ref and surface uncertain publication results

## Tests
- npm run check

Refs #118


VertigoRay:
Implemented the mobile Git-over-File-Station fetch slice on this PR.

What changed:
- Mobile now treats File Station as the Git byte transport: read `HEAD`, resolve `refs/heads/<branch>` or `packed-refs`, then fetch only reachable loose commit/tree/blob objects from the branch tip.
- Removed the previous whole-bare-repo mirror guard/path for mobile.
- Added validated loose-object cache under `.obsidian/plugins/synology-sync/git-cache/v1`; repeat syncs still re-read `HEAD`/ref but skip object downloads on valid Git-oid cache hits.
- Required packed-only objects now fail closed with a clear desktop/bootstrap/pack-handling-needed message until mobile pack reads land.

Validation: `npm run check` passed locally: 14 suites / 209 tests, lint, production build.

Boundary: `SYNO.FileStation.MD5` is not needed as the loose Git-object freshness marker because Git's object SHA/path is already the content-addressed marker. MD5 may still make sense later for mutable normal-file transport metadata or pack-cache validation.

PR #116, opened 2026-05-27
Simplify mobile cache architecture notes

VertigoRay:
## Summary
- Simplifies the mobile Git cache/staging architecture section.
- Removes implementation-specific storage/tooling details from the architecture guidance.
- Keeps the core product model: plugin-owned cache/staging, separate NAS lease vs local cache marker vs live-vault snapshot/diff, safe rebuild path, and cache exclusion from user sync content.

## Validation
- Docs-only change; reviewed diff locally.

PR #128, opened 2026-05-30
Release 2026.0530.1 Git-over-File-Station recovery

### LINKED ISSUE #127: Fix repeated first-sync behavior after completed Git-over-File-Station sync
## Problem
VertigoWerk is showing repeated first-sync/materialization behavior on subsequent syncs. Daily notes can be preserved as conflict copies even though the vault has already completed Git-over-File-Station sync.

## Evidence
- Windows VertigoWerk has a plugin cache marker at ref `5e8e31884198de3efd821dfe42a0ebf3c7df47df` with `completeness=partial` and `reason=sync completed`.
- The live vault `.git` only tracks a handful of setup/plugin files, so it is not a durable Git-over-File-Station baseline.
- The mobile/File Station sync engine uses an in-memory repo. If that repo starts without local commits on each run, the sync can re-enter the first-sync/non-empty-vault path.
- The existing first-sync path preserves same-path local/remote differences as conflict copies rather than running established-history Git merge behavior.

## Desired behavior
- Once a vault has completed a Git-over-File-Station sync for a remote ref, later syncs should not treat the vault as a brand-new first sync solely because the in-memory Git repo starts empty.
- Repeat sync should reconstruct or persist enough plugin-owned baseline state to enter normal Git merge/reconcile behavior.
- Conflict artifacts should only be created for real unmerged Git conflicts or explicit first-sync policy outcomes.
- Logs should clearly state why the engine chose first-sync vs established-history behavior, including cache marker/ref, runtime identity, expected/actual remote ref, and publish/upload outcomes.

## Acceptance
- Add a regression test where a prior cache marker exists, the remote ref matches the marker, local Markdown changed, and the engine does not create repeated first-sync conflict copies.
- Add publish diagnostics that show object/ref upload success, skip reason, or failure.
- Keep behavior aligned with ARCHITECTURE.md; do not change the architecture doc unless the implemented behavior intentionally changes it.

### LINKED ISSUE #129: Implement persistent plugin-owned mobile Git state
## Problem
The current Git-over-File-Station mobile engine still rebuilds most Git state inside an in-memory filesystem each run. PR #128 adds a safety recovery path using the existing cache marker, but it does not replace the underlying in-memory Git state model.

Ray challenged the assumption that mobile cannot persist Git metadata. ARCHITECTURE.md already names the preferred long-term model: a plugin-owned persistent Git cache/staging area, not a literal vault-root `.git`, so repeated syncs can reuse validated state and avoid rebuilding/materializing the world.

## Desired design
Investigate and implement a persistent plugin-owned Git state directory/store, for example:

```text
.obsidian/plugins/synology-sync/git-state/
```

or another Obsidian-mobile-safe storage backend such as LightningFS/IndexedDB, with strict exclusion from user sync content.

## Requirements
- Do not create a literal `.git` at the vault root.
- Keep Git metadata under plugin-owned storage or equivalent mobile-safe persistent storage.
- Exclude that storage from vault sync, Obsidian config sync, indexing/workspace state, staging, and commits.
- Support isomorphic-git `dir`/`gitdir` separation or equivalent so live vault files and Git metadata are not conflated.
- Snapshot the readable vault before remote materialization, per ARCHITECTURE.md.
- Validate cache marker/ref/object completeness under the NAS lease before reusing state.
- Provide clear logs showing Git state location/backend, cache completeness, ref, and rebuild reason.
- Provide a user-visible clear/rebuild-cache path.
- Fail closed on quota/eviction/corruption/partial writes.

## Acceptance
- A second mobile sync after successful first sync reuses persistent Git state and does not need to reconstruct HEAD/index/object state from scratch.
- Local edits after a successful sync enter normal established-history merge/commit/publish behavior without relying only on the PR #128 cache-marker recovery fallback.
- No plugin-owned Git metadata appears in `git.listFiles()` for the user vault or gets uploaded to the NAS as vault content.
- Tests cover cache reuse, corrupt cache invalidation, and storage missing/evicted fallback.

Related: #127, #128.

### LINKED ISSUE #130: Abort safely when notes change during Git-over-File-Station sync
## Problem
Ray reproduced edit loss/conflicts when a note is edited while sync is in progress:

1. Start editing a note.
2. Sync starts.
3. Continue editing the same note while sync is running.
4. Conflict artifacts appear and some edits are lost.
5. Repeating the pattern can repeat the damage.

This is not only the repeated-first-sync marker bug. It is a live-vault snapshot stability problem: the sync can snapshot/read one version of a note, continue remote/git work, then publish or apply checkout changes after the user has already changed the same note again.

## Architecture alignment
ARCHITECTURE.md requires a pre-materialization live-vault snapshot and says that on non-atomic runtimes the engine should verify the snapshot did not change; if files change while a commit is being built, abort or defer that file to a later sync rather than committing an inconsistent snapshot.

## Desired behavior
- If any sync-managed live vault file changes after the sync snapshot and before publish/apply, abort/defer safely.
- Do not publish stale bytes for a note that changed mid-sync.
- Do not overwrite a note that changed mid-sync while applying checkout/materialization.
- Log the exact file(s) and say the user should rerun sync after editing settles.

## Acceptance
- Add a regression test that mutates a mocked vault file after snapshot but before publish/apply and asserts no publish occurs.
- Add a regression test that applyCheckoutChanges does not overwrite a file whose live vault hash changed after the snapshot.
- The error should be actionable and should not create conflict copies for this transient editing race.

Related: #127, #128.

VertigoRay:
## Summary
- recover established Git-over-File-Station checkouts from plugin-owned mobile Git state when the in-memory repo starts empty
- persist validated Git metadata under `.obsidian/plugins/synology-sync/git-cache/v1/state` with manifest fingerprints, safe rebuild on corrupt/evicted state, and no vault-root `.git`
- stop stale live-edit publishes/applies when notes change mid-sync
- avoid repeated first-sync conflict copies and skip re-uploading objects already reachable from the expected old File Station ref
- fold in the mobile cache architecture cleanup and release metadata for `2026.0530.1`

Fixes #127.
Fixes #129.
Fixes #130.
Includes #116.

## Validation
- `npm.cmd test -- --runInBand tests/git-filestation-mobile.spec.ts`
- `npm.cmd run check` (14 suites / 226 tests / lint / production build)


VertigoRay:
Follow-up from Ray's Discord challenge: PR #128 is a tactical safety fix for the observed repeated-first-sync/conflict-copy behavior, not the final mobile Git storage architecture.

ARCHITECTURE.md already points to the better long-term model: persistent plugin-owned Git cache/staging, not a literal vault-root `.git`, so mobile can reuse validated Git metadata across runs. This PR keeps the live behavior safer by recovering established checkout state from a matching marker/ref, while the follow-up should move mobile off the rebuild-every-run in-memory model.

VertigoRay:
Added commit `46268df` for Ray's live-edit repro:

- Rechecks synced live-vault files after the snapshot and before local commit, publish, and checkout apply.
- If a note changed while sync was running, sync now aborts before publishing stale bytes or overwriting the live file, logs the exact path, and asks for a rerun after editing settles.
- Added regression coverage that mutates `Daily/live-edit.md` after the snapshot and asserts no object/ref publish occurs.

Validation on Windows local repo:
- `npm.cmd test -- --runInBand` -> 222 passed
- `npm.cmd run lint` -> passed
- `npm.cmd run build` -> passed

This keeps PR #128 as the tactical safety/release PR. The longer-term ARCHITECTURE-aligned persistent plugin-owned Git state remains tracked separately in #129.

VertigoRay:
Follow-up from the 18:04 VertigoWerk log:

- The NAS lease was not an orphan: `/home/Obsidian/git/VertigoWerk.git/.synology-sync/leases/main.lock/lease.json` existed.
- It was owned by this same VertigoWerk Windows sync identity, created `2026-05-30T17:51:11Z`, expiring `2026-05-30T18:21:11Z`.
- I cleared that stale self-lock after verifying the 18:04 run failed immediately and the lock owner matched this device.
- Added commit `d848271` so future lease-held errors include held lease metadata: this/other owner, created time, and expiry time.

Validation after the diagnostic patch:
- `npm.cmd test -- --runInBand` -> 223 passed
- `npm.cmd run lint` -> passed
- `npm.cmd run build` -> passed

Also refreshed the local VertigoWerk installed plugin bundle from this PR build.

VertigoRay:
Smoke test from live Obsidian VertigoWerk run:

- Current Obsidian sync acquired a fresh NAS lease at `2026-05-30T18:19:33Z`, owned by the same VertigoWerk Windows sync identity, expiring `18:49:33Z`.
- As of `18:38Z`, `Synology Sync Logs/latest-run.md` had not updated since the earlier 18:04 stale-lock failure.
- Remote `refs/heads/main` was still `6e0c3e494c46...`, matching the lease expected old ref, so nothing had published yet.
- Obsidian renderer remained responsive but kept accumulating CPU, which makes this a failed smoke for runtime progress/completion, not a clean pass.

Follow-up pushed in commit `35ae82f`:
- Persist `latest-run.md` at `started`, every 15s as `running`, and `finished` so long syncs are no longer black boxes.
- This note path is already excluded from Git sync (`Synology Sync Logs/latest-run*.md`), so progress writes should not become synced content or trigger live-edit stale-write guards.

Validation after the patch:
- `npm.cmd test -- --runInBand` -> 223 passed
- `npm.cmd run lint` -> passed
- `npm.cmd run build` -> passed

I copied the new bundle into the local VertigoWerk plugin folder, but the currently running Obsidian process must reload/restart before this progress-log patch executes.

